PR #183, opened 2014-11-21
Implement BuildImage cancellation (incomplete)

### LINKED ISSUE #182: No way to cancel a build
Builds can be expensive and it can be that a build in process in no longer necessary, for example if a container description is superseded in quick succession.

Can we add a `chan<- struct{}` to BuildImageOptions which, if closed, signals that the build should be cancelled?

I'm not sure about how the docker API works, but I would hope that the build is cancelled if the connection to the docker server is closed.

pwaller:
This prototypes BuildImage cancellation.

It's split into three commits which should make the implementation clear.

Unfortunately it doesn't work for now until docker itself implements cancellation.

I've supplied this pull request to document the mechanics of this since I had already done a fair amount of the implementation on the assumption it would work before discovering that docker itself doesn't support it.

Therefore, I will immediately close this pull request and hopefully we can reopen it and revisit it in the future.

Fixes #182

### DIFF
--- a/image.go
+++ b/image.go
@@ -341,6 +341,8 @@ type BuildImageOptions struct {
 	Auth                AuthConfiguration  `qs:"-"` // for older docker X-Registry-Auth header
 	AuthConfigs         AuthConfigurations `qs:"-"` // for newer docker X-Registry-Config header
 	ContextDir          string             `qs:"-"`
+	// User controlled channel to allow cancellation
+	Cancel <-chan struct{}
 }
 
 // BuildImage builds an image from a tarball's url or a Dockerfile in the input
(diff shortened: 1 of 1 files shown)


pwaller:
Docker may soon gain build cancellation. Please see https://github.com/docker/docker/issues/9774. @fsouza would you be willing to reopen this pull request and review at this point?

fsouza:
Hey @pwaller, thanks for contributing and raising this again, I will have a look in the code soon, but it looks interesting!

fsouza:
BTW, could you create another pull request? Thanks!

pwaller:
Resubmitted as #208

PR #193, opened 2014-12-30
client: add TLS support, fixes #184

### LINKED ISSUE #184: AttachToContainer should support TLS
[AttachToContainer][meth] doesn't currently [support TLS][line]

  [meth]: https://github.com/fsouza/go-dockerclient/blob/master/container.go#L652
  [line]: https://github.com/fsouza/go-dockerclient/blob/master/client.go#L463

fatih:
Without this we cannot use authenticated connection over TCP, only UNIX
socket is currently supported.

### DIFF
--- a/client.go
+++ b/client.go
@@ -460,30 +460,49 @@ func (c *Client) hijack(method, path string, success chan struct{}, setRawTermin
 		protocol = "tcp"
 		address = c.endpointURL.Host
 	}
-	dial, err := net.Dial(protocol, address)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


fatih:
Any feedback on this? Thanks.

fsouza:
Hey @fatih, these changes look good, thank you! I will do some minor cleanup in blank lines and too simple comments just after merging your code.

Sorry for the huge delay, I was kinda offline these days, but I will try to keep up with changes now!

PR #194, opened 2015-01-06
Excludes is not part of docker/pkg/archive

apriendeau:
Running `go get github.com/fsouza/go-dockerclient`

returns

github.com/fsouza/go-dockerclient
../../../github.com/fsouza/go-dockerclient/tar.go:30: unknown archive.TarOptions field 'Excludes' in struct literal

This updates with the code docker archive Library


asim:
The build fails on Travis because you did not run go fmt.

PR #198, opened 2015-01-07
README: ListImages accept only ListImageOptions

artem-sidorenko:
Otherwise you get 

    ./didep.go:24: cannot use true (type bool) as type docker.ListImagesOptions in function argument


fsouza:
Thank you!

PR #204, opened 2015-01-19
Use -f flag in go get for golint

sweeney:
- forces `go get -u` not to verify that each package has been checked out from the source control repository implied by its import path


fsouza:
Thank you!

Issue #203, opened 2015-01-13
Dead lock in docker events


gambol99:
I keep hitting a dead lock when listening to docker events; a section of the code below

	   docker_events := make(chan *docker.APIEvents)
	    	go func() {
	    		/* step: add our channel as an event listener for docker events */
	    		if err := r.Docker.AddEventListener(docker_events); err != nil {
	    			glog.Fatalf("Unable to register docker events listener, error: %s", err)
	    			return
	    		}
	    		/* step: start the event loop and wait for docker events */
	    		glog.V(5).Infof("Entering into the docker events loop")
	    		for {
	    			select {
	    			case event := <-docker_events:
	    				glog.V(4).Infof("Received docker event status: %s, id: %s", event.Status, event.ID)
	    				switch event.Status {
	    				case DOCKER_START:
	    					go r.ProcessDockerCreation(event.ID, channel)
	    				case DOCKER_DESTROY:
	    					go r.ProcessDockerDestroy(event.ID, channel)
	    				}
	    			}
	    		}
	    		glog.Errorf("Exitting the docker events loop")
	    	}()

Eventually the loop will stop listening to docker events - performing a kill -6 PID

    goroutine 22 [semacquire]:
    sync.runtime_Semacquire(0xc20805215c)
      /usr/lib/golang/src/pkg/runtime/sema.goc:199 +0x30
    sync.(*RWMutex).RLock(0xc208052150)
      /usr/lib/golang/src/pkg/sync/rwmutex.go:36 +0x5e
    github.com/fsouza/go-dockerclient.(*eventMonitoringState).isEnabled(0xc208052150, 0xc20815c400)
      /home/jest/go/src/github.com/gambol99/embassy/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/event.go:219 +0x33
    github.com/fsouza/go-dockerclient.(*eventMonitoringState).monitorEvents(0xc208052150, 0xc2080a6000)
      /home/jest/go/src/github.com/gambol99/embassy/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/event.go:173 +0xf7
    created by github.com/fsouza/go-dockerclient.(*eventMonitoringState).enableEventMonitoring
      /home/jest/go/src/github.com/gambol99/embassy/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/event.go:148 +0x113
    
    goroutine 36 [semacquire]:
    sync.runtime_Semacquire(0xc20805215c)
[...]

PR #206, opened 2015-01-23
Deadlock workaround

blalor:
Attempting workaround for deadlock reported in https://github.com/fsouza/go-dockerclient/issues/203; no need for updateLastSeen to be called from a goroutine.

I've used this in my [fork of registrator](https://github.com/blalor/registrator) for a few days and I haven't seen the deadlock that @gambol99 and I have experienced.  After looking through the code, I see no need for `eventState.updateLastSeen` to be called in a goroutine.  I also thought it more appropriate that that be handled before dispatching the event.

### DIFF
--- a/event.go
+++ b/event.go
@@ -182,8 +182,8 @@ func (eventState *eventMonitoringState) monitorEvents(c *Client) {
 				eventState.terminate()
 				return
 			}
+			eventState.updateLastSeen(ev)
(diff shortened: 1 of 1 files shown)


fsouza:
Hey @blalor, thanks for contributing. That's definitely a good solution and it indeed makes sense. I'm merging this issue, and closing #203.

PR #208, opened 2015-01-27
Implement BuildImage cancellation

### LINKED ISSUE #182: No way to cancel a build

pwaller:
This prototypes BuildImage cancellation.

It's split into three commits which should make the implementation clear.

Unfortunately it doesn't work for now until docker itself implements cancellation.

Fixes #182

See https://github.com/docker/docker/pull/9774

This feature will not work until https://github.com/docker/docker/pull/9774 or something like it is merged.

### DIFF
--- a/client.go
+++ b/client.go
@@ -339,7 +339,7 @@ func (c *Client) do(method, path string, data interface{}, forceJSON bool) ([]by
 	return body, resp.StatusCode, nil
 }
 
-func (c *Client) stream(method, path string, setRawTerminal, rawJSONStream bool, headers map[string]string, in io.Reader, stdout, stderr io.Writer) error {
+func (c *Client) stream(method, path string, setRawTerminal, rawJSONStream bool, headers map[string]string, in io.Reader, stdout, stderr io.Writer, cancelled <-chan struct{}) error {
 	if (method == "POST" || method == "PUT") && in == nil {
 		in = bytes.NewReader(nil)
 	}
--- a/client.go
+++ b/client.go
@@ -35,6 +35,11 @@ var (
 	ErrConnectionRefused = errors.New("cannot connect to Docker endpoint")
 
 	apiVersion112, _ = NewAPIVersion("1.12")
+	
+	// ErrCancelled is returned when the API-user requests an operation is
+	// cancelled.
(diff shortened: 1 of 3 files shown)


kidoman:
Should we look at this as a way to expose this functionality?

https://godoc.org/golang.org/x/net/context#WithCancel

pwaller:
docker/docker#9774 got merged, so we should revisit this some time.

fsouza:
Hi @pwaller, care to rebase the pull request? Thank you!

pwaller:
I've rebased. I'd love to see this work via the [`net.Context`](https://godoc.org/golang.org/x/net/context#Context) mechanism though. I'd like to rework it to use that, but I don't have a lot of time right at this moment.

pwaller:
Note for anyone taking this up (and a reminder to myself in case that is me): if you want to cancel a request it is racy if you're not careful. You can only cancel a request once it is in flight - if you cancel it before it is in flight... it won't work.

[So you have to take some approach that looks like this](https://github.com/golang/go/commit/ececbe89d4d5ab151333cbadae725619c0ad9dd8). I.e, you can only fire off [the goroutine responsible for performing the cancel](https://github.com/golang/go/commit/ececbe89d4d5ab151333cbadae725619c0ad9dd8#diff-d863507a61be206d112f6e00e6d812a2R144) once you've seen the first `Read()` call on the request body. Yick, but that's how it is.

pwaller:
So I'm tempted to close this pull request and start afresh, with that in mind. But I don't have much time right now.

If I were doing it, I would use `x/net/context.Context`, and allow it to be passed from the clients via the `FooOpts` structs. If it were zero, it would take the default (background) context.

pwaller:
I'm going to relinquish my claim to implementing this for now, but I'm willing to give a hand with the review if you ping me. If I feel a pressing need for it again I'll come back and do it then.

PR #210, opened 2015-02-06
Add support for SecurityOpts in Config

mmateliscoursera:
Support the SecurityOpts option in Config, see: https://docs.docker.com/reference/api/docker_remote_api_v1.16/ for context.


fsouza:
Thanks!

PR #211, opened 2015-02-11
Fix deadlock bug in event.go:

burke:
`noListeners` claims `eventState` lock, which is already held in one invocation. Without the `RLock` code, the method is simple enough that inlining felt like the right answer.


fsouza:
Nice catch, thanks!

PR #216, opened 2015-02-16
Add 'dockerfile' option to image build, according to docker api v1.17

ybogdanov:
Hey,

In the version 1.5 of Docker they added support to specify custom Dockerfile target for builds. This is must have feature in our pipeline. See http://docs.docker.com/reference/api/docker_remote_api_v1.17/#build-image-from-a-dockerfile

Added this option to `BuildImageOptions`, tested, it works.


fsouza:
Hey @ybogdanov, thanks for contributing. This change looks good.

Issue #180, opened 2014-11-12
Versions


mcuadros:
I am using this library at my project [dockership](https://github.com/mcuadros/dockership), since this commit: https://github.com/fsouza/go-dockerclient/commit/90496c840a3a13cc9976af74a375aef57daaa1de, `docker.PortBinding.HostIp` has changed to `docker.PortBinding.HostIP`

I suggest use versioning in this type of changes because is breaking the compilation of 3rd party projects.

Maybe http://labix.org/gopkg.in could be a good solution.

Best

budhrg:
Hi,

Any update on it?

Thanks.

fsouza:
Hi @budhrg, I will start working on the plans described here in this weekend, thanks for your interest in go-dockerclient!

fsouza:
Same as #191. Let's see what are the next steps in Go versioning/package management.

PR #217, opened 2015-02-16
Add RemoveImageExtended() to support 'force' param for removing images

ybogdanov:
Hi,

There is an option `force` to remove images even if there is a child container existing. http://docs.docker.com/reference/api/docker_remote_api_v1.17/#remove-an-image

I created a separate function
`func (c *Client) RemoveImageExtended(name string, opts RemoveImageOptions) error`
without breaking existing `RemoveImage` function.

What do you think? Should it be merged into existing function?

### DIFF
--- a/image.go
+++ b/image.go
@@ -146,8 +146,8 @@ func (c *Client) RemoveImage(name string) error {
 //
 // See http://goo.gl/6V48bF for more details.
 type RemoveImageOptions struct {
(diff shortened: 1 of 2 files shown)


fsouza:
I really appreciate this approach, let's go with this one and later unify the method as part of the issue #180.

Thank you!

PR #219, opened 2015-02-16
Fix spelling of NewVersionedTLSClient with depricated original function

nickethier:
This is a really small bug, but it gave me a headache until I finally copy/paste the method from godoc.

This is a proposal to deprecate the existing `NewVersionnedTLSClient` and point it to the renamed function `NewVersionedTLSClient` which is consistent with the spelling of `NewVersionedClient`.

Cheers!

### DIFF
--- a/client.go
+++ b/client.go
@@ -174,7 +174,7 @@ func NewVersionedClient(endpoint string, apiVersionString string) (*Client, erro
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


fsouza:
Ugh, that's a shame. Sorry for the bad review in the original change.

I will keep the deprecated version for a while and then delete it, probably as part of the fix for #180.

Thanks for contributing, and for your interest in go-dockerclient!

Issue #129, opened 2014-07-30
json decode memory failed


CMGS:
the docker api will return memory config like "Memory": 6.5536e+07 in json. and it will make unmarshal failed because config.Memory is uint64 type.

example: 
`json: cannot unmarshal number 5.36870912e+09 into Go value of type uint64`

PR #224, opened 2015-02-28
Container Config struct Memory should be float64 #111

Hevienz:
This has been fixed in #111, but it appears again.
https://github.com/fsouza/go-dockerclient/pull/111


fsouza:
Please have a look at #129, #131 and GoogleCloudPlatform/kubernetes#589. Using float64 for memory was actually a Docker 1.0 bug, rounding the amount of memory is no good, and they have fixed it already.

PR #225, opened 2015-03-02
Support for stats endpoint introduced in 1.5

Tonnu:
This PR implements the necessary changes to support collecting data from the /stats/ endpoint. 

It includes a new method in client.go which handles the request with stream(). The stream() method itself has also has been updated to include two extra parameters "messages" and "statsBool". 

Messages is a channel which can be optionally passed if the user wishes to do something different with the gathered metrics on the client side. statsBool should be set to true if the user is accessing the /stats/ endpoint. 

Finally it includes a new method decodeStats() which handles the actual unmarshalling of the data into the newly added structs.


Tonnu:
Cleaner implementation described in https://github.com/fsouza/go-dockerclient/pull/232.

PR #230, opened 2015-03-06
Correct Volumes type for container config

azoff:
`Config.Volumes` is a `map[string]stuct{}` but `Container.Volumes` is a `map[string]string`.

This patch makes them match.

### DIFF
--- a/container.go
+++ b/container.go
@@ -169,32 +169,32 @@ func parsePort(rawPort string) (int, error) {
 // Config does not the options that are specific to starting a container on a
 // given host.  Those are contained in HostConfig
 type Config struct {
-	Hostname        string              `json:"Hostname,omitempty" yaml:"Hostname,omitempty"`
-	Domainname      string              `json:"Domainname,omitempty" yaml:"Domainname,omitempty"`
-	User            string              `json:"User,omitempty" yaml:"User,omitempty"`
-	Memory          int64               `json:"Memory,omitempty" yaml:"Memory,omitempty"`
-	MemorySwap      int64               `json:"MemorySwap,omitempty" yaml:"MemorySwap,omitempty"`
-	CPUShares       int64               `json:"CpuShares,omitempty" yaml:"CpuShares,omitempty"`
-	CPUSet          string              `json:"Cpuset,omitempty" yaml:"Cpuset,omitempty"`
-	AttachStdin     bool                `json:"AttachStdin,omitempty" yaml:"AttachStdin,omitempty"`
-	AttachStdout    bool                `json:"AttachStdout,omitempty" yaml:"AttachStdout,omitempty"`
-	AttachStderr    bool                `json:"AttachStderr,omitempty" yaml:"AttachStderr,omitempty"`
-	PortSpecs       []string            `json:"PortSpecs,omitempty" yaml:"PortSpecs,omitempty"`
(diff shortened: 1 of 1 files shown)


container.go
```diff
@@ -169,32 +169,32 @@ func parsePort(rawPort string) (int, error) {
...
+	PortSpecs       []string          `json:"PortSpecs,omitempty" yaml:"PortSpecs,omitempty"`
+	ExposedPorts    map[Port]struct{} `json:"ExposedPorts,omitempty" yaml:"ExposedPorts,omitempty"`
+	Tty             bool              `json:"Tty,omitempty" yaml:"Tty,omitempty"`
+	OpenStdin       bool              `json:"OpenStdin,omitempty" yaml:"OpenStdin,omitempty"`
+	StdinOnce       bool              `json:"StdinOnce,omitempty" yaml:"StdinOnce,omitempty"`
+	Env             []string          `json:"Env,omitempty" yaml:"Env,omitempty"`
+	Cmd             []string          `json:"Cmd,omitempty" yaml:"Cmd,omitempty"`
+	DNS             []string          `json:"Dns,omitempty" yaml:"Dns,omitempty"` // For Docker API v1.9 and below only
+	Image           string            `json:"Image,omitempty" yaml:"Image,omitempty"`
+	Volumes         map[string]string `json:"Volumes,omitempty" yaml:"Volumes,omitempty"`
```
azoff: this is the actual change, all other lines are a result of `go fmt`

fsouza:
Hi @azoff, this is intentional, because in the Docker API they don't actually match:

- http://docs.docker.com/reference/api/docker_remote_api_v1.17/#create-a-container
- http://docs.docker.com/reference/api/docker_remote_api_v1.17/#inspect-a-container

So using map[string]string in Config would cause an unmarshalling error in the Docker API.

Thanks for contributing!

azoff:
fair enough, odd that it doesn’t map to the -v switch in the CLI, which I’d assume is a string/string

--  
Jonathan Azoff
Sent with Sparrow (http://www.sparrowmailapp.com/?sig)

fsouza:
@azoff Docker uses ``-v`` in HostConfig.Binds parameter, and it's a slice of strings. The value of the flag also used in the Volumes, and it's map[string]struct{} so Docker can ensure that there are no duplicate volumes.

azoff:
Ahhh! That makes a little more sense :) Thanks for clarifying that!

--  
Jonathan Azoff
Sent with Sparrow (http://www.sparrowmailapp.com/?sig)

fsouza:
No problem, please feel free to open new pull requests or issues in the future.

Thank you!

azoff:
I actually do have one more question, since you're in the giving mood :)

Is there a Config or HostConfig property that correlates with the `ADD` directive from the `Dockerfile` ?

fsouza:
Nope, Config and HostConfig are used for containers, not images. You can use the Dockerfile with the [BuildImage](http://godoc.org/github.com/fsouza/go-dockerclient#Client.BuildImage) method. Currently, Docker doesn't expose the ADD functionality in the remote API.

Sorry for the delay.

fsouza:
I'm sorry, Config and HostConfig are saved in the image when you commit the container, so my first statement is not true. Sorry about the confusion.

PR #231, opened 2015-03-07
Add support for tag in DockerServer.pullImage (testing)

vinceprignano:
If tag is set in URL Query, `pullImage` should save `fromImageName` + `tag`.


fsouza:
Thank you!

PR #232, opened 2015-03-08
Cleaner implementation of 1.5 stats support

Tonnu:
Cleaner channel-based implementation of https://github.com/fsouza/go-dockerclient/pull/225. Stream() method is left untouched and the new functionality is now in the decodeStats() and StatsContainer()

### DIFF
--- a/container.go
+++ b/container.go
@@ -483,52 +483,52 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
 }
 
 type NetworkStats struct {
-	RX_dropped uint64 `json:"rx_dropped"`
-	RX_bytes   uint64 `json:"rx_bytes"`
-	RX_errors  uint64 `json:"rx_errors"`
-	TX_packets uint64 `json:"tx_packets"`
-	TX_dropped uint64 `json:"tx_dropped"`
-	RX_packets uint64 `json:"rx_packets"`
-	TX_errors  uint64 `json:"tx_errors"`
-	TX_bytes   uint64 `json:"tx_bytes"`
+	RX_dropped int64 `json:"rx_dropped"`
+	RX_bytes   int64 `json:"rx_bytes"`
+	RX_errors  int64 `json:"rx_errors"`
+	TX_packets int64 `json:"tx_packets"`
+	TX_dropped int64 `json:"tx_dropped"`
+	RX_packets int64 `json:"rx_packets"`
+	TX_errors  int64 `json:"tx_errors"`
+	TX_bytes   int64 `json:"tx_bytes"`
 }
 
 type MemoryStats struct {
 	Stats     DetailedMemoryStats `json:"stats"`
-	Max_usage uint64              `json:"max_usage"`
-	Usage     uint64              `json:"usage"`
-	Failcnt   uint64              `json:"failcnt"`
-	Limit     uint64              `json:"limit"`
+	Max_usage int64               `json:"max_usage"`
+	Usage     int64               `json:"usage"`
+	Failcnt   int64               `json:"failcnt"`
+	Limit     int64               `json:"limit"`
 }
 
 type DetailedMemoryStats struct {
-	Total_pgmajfault          uint64 `json:"total_pgmajfault"`
-	Cache                     uint64 `json:"cache"`
-	Mapped_file               uint64 `json:"mapped_file"`
-	Total_inactive_file       uint64 `json:"total_inactive_file""`
-	Pgpgout                   uint64 `json:"pgpgout"`
-	Rss                       uint64 `json:"rss"`
-	Total_mapped_file         uint64 `json:"total_mapped_file"`
-	Writeback                 uint64 `json:"writeback"`
-	Unevictable               uint64 `json:"unevictable"`
-	Pgpgin                    uint64 `json:"pgpgin"`
-	Total_unevictable         uint64 `json:"total_unevictable"`
-	Pgmajfault                uint64 `json:"pgmajfault"`
-	Total_rss                 uint64 `json:"total_rss"`
-	Total_rss_huge            uint64 `json:"total_rss_huge"`
-	Total_writeback           uint64 `json:"total_writeback"`
-	Total_inactive_anon       uint64 `json:"total_inactive_anon"`
-	Rss_huge                  uint64 `json:"rss_huge"`
-	Hierarchical_memory_limit uint64 `json:"hierarchical_memory_limit"`
-	Total_pgfault             uint64 `json:"total_pgfault"`
-	Total_active_file         uint64 `json:"total_active_file"`
-	Total_pgpgout             uint64 `json:"total_pgpgout"`
-	Total_cache               uint64 `json:"total_cache"`
-	Inactive_anon             uint64 `json:"inactive_anon"`
-	Active_file               uint64 `json:"active_file"`
-	Pgfault                   uint64 `json:"pgfault"`
-	Inactive_file             uint64 `json:"inactive_file"`
-	Total_pgpgin              uint64 `json:"pgpgin"`
+	Total_pgmajfault          int64 `json:"total_pgmajfault"`
+	Cache                     int64 `json:"cache"`
+	Mapped_file               int64 `json:"mapped_file"`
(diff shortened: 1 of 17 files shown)


fsouza:
Hi @Tonnu, thank you very much for contributing, I had a quick look through the code and it looks good so far. I'm missing some tests, can you add them? Also, I will make some specific comments in the code, most related to code style, and run some tests locally.

container.go
```diff
@@ -482,6 +482,113 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+// NetworkStats describes the data related to the container's memory usage.
+type NetworkStats struct {
+	RXdropped uint64 `json:"rx_dropped"`
+	RXbytes   uint64 `json:"rx_bytes"`
+	RXerrors  uint64 `json:"rx_errors"`
+	TXpackets uint64 `json:"tx_packets"`
+	TXdropped uint64 `json:"tx_dropped"`
+	RXpackets uint64 `json:"rx_packets"`
+	TXerrors  uint64 `json:"tx_errors"`
+	TXbytes   uint64 `json:"tx_bytes"`
```
fsouza: Please use RXDropped, RXBytes, RXErrors, etc. When connecting words, the second words must come with the first letter in uppercase.

container.go
```diff
@@ -482,6 +482,113 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+}
+
+// MemoryStats describes the data realted to the container's memory usage.
+type MemoryStats struct {
+	Stats           DetailedMemoryStats `json:"stats"`
+	MaxUsage        uint64              `json:"max_usage"`
+	Usage           uint64              `json:"usage"`
+	Failcnt         uint64              `json:"failcnt"`
+	Limit           uint64              `json:"limit"`
+	PercentageInUse float64
```
fsouza: Why is this field present and not being used anywhere?

container.go
```diff
@@ -482,6 +482,113 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+	Pgfault                 uint64 `json:"pgfault"`
+	InactiveFile            uint64 `json:"inactive_file"`
+	TotalPgpgin             uint64 `json:"pgpgin"`
+}
+
+// CPUStats describes the data related to the container's CPU usage.
+type CPUStats struct {
+	CPUUsage        CPUUsage `json:"cpu_usage"`
+	SystemCPUUsage  uint64   `json:"system_cpu_usage"`
+	PercentageInUse float64
```
fsouza: Same as MemoryStats.

container.go
```diff
@@ -482,6 +482,113 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+}
+
+// CPUUsage describes the data related to the container's CPU usage.
+type CPUUsage struct {
+	PercpuUsage       []uint64 `json:"percpu_usage"`
+	UsageInUsermode   uint64   `json:"usage_in_usermode"`
+	TotalUsage        uint64   `json:"total_usage"`
+	UsageInKernelmode uint64   `json:"usage_in_kernelmode"`
```
fsouza: Please use:

- PerCPUUsage
- UsageInUserMode
- UsageInKernelMode

container.go
```diff
@@ -482,6 +482,113 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+}
+
+// ContainerStats describes the data as it is retrieved from the /stats/ endpoint.
+type ContainerStats struct {
+	Read    string       `json:"read"`
+	Stream  string       `json:"stream,omitempty"`
+	Network NetworkStats `json:"network"`
+	Memory  MemoryStats  `json:"memory_stats"`
+	CPU     CPUStats     `json:"cpu_stats"`
+	EOF     string
```
fsouza: What's this field?

container.go
```diff
@@ -482,6 +482,113 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+			}
+		}
+	}
+	return nil
+}
+
+// StatsContainer gets container stats based on resource usage
+//
+// See http://goo.gl/eY5NRI for more details.
+func (c *Client) StatsContainer(id string, stats chan ContainerStats) error {
```
fsouza: A good practice when receiving a channel is to ensure the intent in this channel, so you could ensure that you're only going to write to the stats chan by declaring it ``stats chan<- ContainerStats``.

container.go
```diff
@@ -482,6 +482,113 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+	Stream  string       `json:"stream,omitempty"`
+	Network NetworkStats `json:"network"`
+	Memory  MemoryStats  `json:"memory_stats"`
+	CPU     CPUStats     `json:"cpu_stats"`
+	EOF     string
+}
+
+// uses a different struct to unmarshal to when we're calling the stats endpoint
+func decodeStats(r io.Reader, stats chan ContainerStats) error {
+	dec := json.NewDecoder(r)
```
fsouza: That's just a question, should we add a ``defer close(stats)`` here, to signal readers that after this function returns, no new messages will be sent in the channel?

container.go
```diff
@@ -482,6 +482,113 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+	HierarchicalMemoryLimit uint64 `json:"hierarchical_memory_limit"`
+	TotalPgfault            uint64 `json:"total_pgfault"`
+	TotalActiveFile         uint64 `json:"total_active_file"`
+	TotalPgpgout            uint64 `json:"total_pgpgout"`
+	TotalCache              uint64 `json:"total_cache"`
+	InactiveAnon            uint64 `json:"inactive_anon"`
+	ActiveFile              uint64 `json:"active_file"`
+	Pgfault                 uint64 `json:"pgfault"`
+	InactiveFile            uint64 `json:"inactive_file"`
+	TotalPgpgin             uint64 `json:"pgpgin"`
```
fsouza: Please use:

- TotalPageMajorFaults instead of TotalPgmajfault
- PageMajorFaults instead of Pgmajfault
- PageOut instead of Pgpgout
- PageIn instead of Pgpgin
- TotalRSS instead of TotalRss
- TotalRSSHuge instead of TotalRssHuge
- RSSHuge instead of RssHuge
- TotalPageFaults instead of TotalPgFault
- TotalPageOut instead of TotalPgpgout
- PageFaults instead of Pgfault
- TotalPageIn instead of TotalPgpgin

vinceprignano:
Is passing a channel to the function really useful in common usage?

Also, the code always writes to the channel _assuming_ that is open. So if the client for some reason closes the channel outside `decodeStats` will panic. Is this correct?

fsouza:
@vinceprignano that's why it's important to explicitly set the intent of the channel when receiving it, so if I pass a channel to a function that intents to write to this channel (because the parameter is ``chan-< type``), I should not close the channel.

But a better solution would be something like:

```
func StatsContainer(opts StatsContainerOptions) (<-chan Stats, error) {
    reader, writer := io.Pipe()
    stats := make(chan Stats)
    go decodeStats(reader, stats)
    if err := c.stream("GET", fmt.Sprintf("/containers/%s/stats", id), true, true, nil, nil, writer, nil); err != nil {
        return nil, err
    }
    return stats, nil
} 
```

And also add ``defer close(stats)`` to the beginning of the ``decodeStats`` method.

What do you think?

vinceprignano:
IMHO when a client calls `ContainerStats(...)` it gets just the stats for that container at this particular moment with no streaming. 

On the other hand we can have a `TailContainerStats` that supports this behavior and that just returns a channel, as in your example. It's common for the client to use this channel in a range loop or a for loop that reads infinitely from the channel.

fsouza:
@vinceprignano I'm sorry, I thought I had answered you already, guess I'm going crazy x)

So, I like your idea, but I would rather add a parameter to StatsContainerOptions (something like a boolean "Follow"). What do you think?

What about you, @Tonnu?

fsouza:
Now that #262 is merged, I'm closing this one. Thanks for contributing!

PR #236, opened 2015-03-09
Add rename container

ibuildthecloud:
Docker 1.5 now supports renaming a container: https://docs.docker.com/reference/api/docker_remote_api_v1.17/#rename-a-container

### DIFF
--- a/container.go
+++ b/container.go
@@ -239,6 +239,25 @@ type Container struct {
 	ExecIDs    []string          `json:"ExecIDs,omitempty" yaml:"ExecIDs,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


fsouza:
Hi @ibuildthecloud, thanks for contributing! Can you add a test and use an struct with the set of options? Just ensuring that we can extend the client in case Docker extends the API.

ibuildthecloud:
Sure no problem.

ibuildthecloud:
@fsouza Updated to use struct and added a test.

fsouza:
Thank you!

PR #237, opened 2015-03-09
Allow https connections to swarm master.

JustAdam:
Quick fix to allow https connections to a swarm-master.

Perhaps the logic for deciding on whether to use https; based on if a tlsconfig has been provided needs to be discussed; so to avoid a potential long list of port numbers.

### DIFF
--- a/client.go
+++ b/client.go
@@ -621,7 +621,7 @@ func parseEndpoint(endpoint string) (*url.URL, error) {
 		}
 
 		number, err := strconv.ParseInt(port, 10, 64)
-		if err == nil && number == 2376 {
+		if err == nil && (number == 2376 || number == 3376) {
 			u.Scheme = "https"
 		} else {
 			u.Scheme = "http"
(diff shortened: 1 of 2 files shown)


fsouza:
Hi @JustAdam, the reason why we accept 2376 as the docker-s port is because this port is actually reserved with IANA. 3376 is not reserved for Swarm master, so go-dockerclient will not infer this from the given port.

You can create the instance of the client using NewTLSClient with the address "https://<your-host>:3376".

Thanks for contributing!

JustAdam:
Thanks @fsouza  I will switch to using https.  The issue came around from docker-machine and swarm which report their addresses via the environment on tcp://[ip]:[port] (Swarm agent is on 2375 and swarm master is on 3376).  If the tcp schema is used it is impossible to get tls working if the port isn't 2376 which is why I suggest perhaps checking if a tlsconfig had been supplied.

fsouza:
This is an interesting approach. If we're calling NewTLSClient, we sould always use the https scheme. Wanna work on that?

Thank you!

fsouza:
Hi @JustAdam, I've just pushed f1f91d5ba55810454f1d75e61d61d8b4c45e6e9b, which implement this behavior.

Thanks for contributing!

JustAdam:
:+1: :)

PR #238, opened 2015-03-09
Added container's swarm node information.

JustAdam:
Updated so InspectContainer also includes Swarm node information.  I appreciate that this isn't apart of the official api, but swarm is becoming/will become a much bigger part of the Docker ecosystem that I think it warrants inclusion.

There may be other Swarm related information about that needs including in other methods, haven't checked fully yet :)

### DIFF
--- a/container.go
+++ b/container.go
@@ -197,6 +197,17 @@ type Config struct {
 	OnBuild         []string            `json:"OnBuild,omitempty" yaml:"OnBuild,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


fsouza:
Hi @JustAdam, does this information come from the Swarm API only?

JustAdam:
Hi @fsouza, looks like it is just the Swarm API for now; https://docs.docker.com/swarm/API/

fsouza:
That's fair, thanks for contributing!

PR #240, opened 2015-03-12
StartContainer - propagate error from docker api (response body)

ybogdanov:
It took some time to figure out why I was getting "container not found" error while trying to start just created container. Docker daemon logs helped - the `CMD` was failing, but its api gives 404 in such case, along with descriptive error still (in http response body).
So it makes sense to propagate this error to `NoSuchContainer` to save time for others.

### DIFF
--- a/container.go
+++ b/container.go
@@ -752,12 +752,10 @@ type NoSuchContainer struct {
 }
 
 func (err *NoSuchContainer) Error() string {
-	switch err.Err.(type) {
-	case error:
-		return "No such container: " + err.ID + " - " + err.Err.Error()
-	default:
-		return "No such container: " + err.ID
+	if err.Err != nil {
+		return err.Err.Error()
 	}
+	return "No such container: " + err.ID
 }
 
 // ContainerAlreadyRunning is the error returned when a given container is
--- a/container.go
+++ b/container.go
@@ -386,7 +386,7 @@ func (c *Client) StartContainer(id string, hostConfig *HostConfig) error {
 	path := "/containers/" + id + "/start"
 	_, status, err := c.do("POST", path, hostConfig, true)
 	if status == http.StatusNotFound {
-		return &NoSuchContainer{ID: id}
+		return &NoSuchContainer{ID: id, Err: err}
 	}
 	if status == http.StatusNotModified {
 		return &ContainerAlreadyRunning{ID: id}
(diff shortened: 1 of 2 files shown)


ybogdanov:
Maybe, it makes sense to do it like Docker did https://github.com/docker/docker/blob/master/api/client/utils.go#L107

fsouza:
Hum, that's interesting. I didn't know that Docker used 404 on this one, nice catch!

container.go
```diff
@@ -747,11 +747,17 @@ func (c *Client) ExportContainer(opts ExportContainerOptions) error {
...
 }
 
 func (err *NoSuchContainer) Error() string {
-	return "No such container: " + err.ID
+	switch err.Err.(type) {
+	case error:
+		return "No such container: " + err.ID + " - " + err.Err.Error()
+	default:
+		return "No such container: " + err.ID
+	}
```
fsouza: Please use:

```
if err.Err != nil {
    return fmt.Sprintf("No such container: %s - %s", err.ID, err.Err)
}
return "No such container: " + err.ID
```

But I'm still thinking whether we should rename the error type.

ybogdanov: Yeah, looks silly now. Look how they did it: https://github.com/docker/docker/blob/master/api/client/utils.go#L112
They have two types of generic errors, one if body present, another if body is empty.
We can do the same check, so if body is empty, we give `NoSuchContainer`, if there is some data in body, give some generic error, putting body there.

fsouza: Yeah, that makes sense. It's good approach. Can you do it?

Thank you!

ybogdanov:
So, `client.go` itself already doint that work for us. It gets http response body and wraps it into `Error` struct. I think it's ok to wrap it with `NoSuchContainer` struct in `container.go` since we've got 404 anyway. However we need to propagate an appropriate error still.

In both cases docker api gives a descriptive message, so we can use that directly:

container not found:
```
API error (404): No such container: sdf
```

CMD failed:
```
API error (404): Cannot start container 4afa230e3c824c90f0de0a5d02794048a233ac1f05e0d8e3076a32b3c9e0e629: exec: "echo Haha > /app/README.md": stat echo Haha > /app/README.md: no such file or directory
```

In case `Err` property wasn't given, we fall back to a default message `"No such container: ID"`.

Looks reasonable to me right now. There also could be other places (errors) where such logic should be applied.

fsouza:
Awesome, I'm merging this pull request then.

Thanks for contributing!

PR #242, opened 2015-03-24
NewClientFromEnv, NewVersionedClientFromEnv mirroring official Docker client

peter-edge:
I've had this in my own code for a while and thought I'd share it across. It's essentially a direct port of https://github.com/docker/compose/blob/master/compose/cli/docker_client.py and I've been using it for a long time for non-TLS and TLS clients.

If it's something you'd accept, I can add tests if necessary, but looking at how the tests are setup now, additional tests may be redundant.

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,6 +20,8 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"os"
+	"path/filepath"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -127,6 +129,61 @@ type Client struct {
...
+// official Docker client, using a specific remote API version.
+func NewVersionedClientFromEnv(apiVersionString string) (*Client, error) {
+	dockerHost := os.Getenv("DOCKER_HOST")
+	if dockerHost == "" {
+		return nil, errors.New("DOCKER_HOST must be set")
+	}
+	if os.Getenv("DOCKER_TLS_VERIFY") == "" {
+		return NewVersionedClient(dockerHost, apiVersionString)
+	}
+	parts := strings.SplitN(dockerHost, "://", 2)
```
peter-edge: This may be redundant from https://github.com/fsouza/go-dockerclient/commit/f1f91d5ba55810454f1d75e61d61d8b4c45e6e9b

client.go
```diff
@@ -127,6 +129,61 @@ type Client struct {
...
+	dockerHost = fmt.Sprintf("https://%s", parts[1])
+	dockerCertPath := os.Getenv("DOCKER_CERT_PATH")
+	if dockerCertPath == "" {
+		home := os.Getenv("HOME")
+		if home == "" {
+			return nil, errors.New("HOME must be set if DOCKER_TLS_VERIFY is set and DOCKER_CERT_PATH is not set")
+		}
+		dockerCertPath = filepath.Join(home, ".docker")
+	}
+	if err := checkFileExists(dockerCertPath); err != nil {
```
peter-edge: You may do this somewhere else, so this might also be redundant.

PR #248, opened 2015-04-08
Make the first example in the README work

guillermo:
* ListImages no longer receives a ListImagesOption
* TYPO: img.ParentID is img.ParentId


fsouza:
Which version of go-dockerclient you're using?

At master:

* https://github.com/fsouza/go-dockerclient/blob/master/image.go#L100
* https://github.com/fsouza/go-dockerclient/blob/master/image.go#L28

guillermo:
You are right.

I pick the wrong repo.

PR #252, opened 2015-04-10
fix missing word

pweil-:
Just a comment update


fsouza:
Good catch, thank you!

PR #254, opened 2015-04-17
ExportImages API call to export multiple images

peter-edge:


peter-edge:
This API call has actually been around for a while, and makes my work here https://github.com/peter-edge/go-dockerbackup/blob/master/client.go into about 20 lines :)

fsouza:
LGTM

Thank you!

PR #255, opened 2015-04-17
Digests option for ListImages

peter-edge:
This would be my first change for this repository where the option is not backwards-compatible, so I am not sure on your policy here. This was added in 1.18. Let me know :)


fsouza:
Hi @peter-edge, thanks for contributing! This isn't backward incompatible because it doesn't break old code, so it's fine to merge. That's the reason we use structs for specifying function parameters, it's easy to add a new parameter without breaking old code.

PR #256, opened 2015-04-19
Container statistics

peter-edge:
https://docs.docker.com/reference/api/docker_remote_api_v1.18/#get-container-stats-based-on-resource-usage


peter-edge:
I literally just noticed there was a separate pull request for this. My bad. Feel free to take this one or take things from it (the tests, for example).

peter-edge:
I looked through the other pull request, it hasn't been updated since mid-March and I (think) this pull request is more robust, maybe we should go with this one instead? My bad for not noticing this earlier.

I used int64 instead of uint64 to more closely match the rest of the API, but can change this.

peter-edge:
Travis passing.

peter-edge:
Here is an example usage for a non-TLS Docker client. Note that `Stats` only returns when the container is REMOVED, not just exited.

```
package main

import (
	"fmt"
	"os"

	"github.com/peter-edge/go-dockerclient"
)

func main() {
	if err := do(); err != nil {
		fmt.Fprintf(os.Stderr, "%s\n", err.Error())
		os.Exit(1)
	}
	fmt.Fprintf(os.Stdout, "ok\n")
	os.Exit(0)
}

func do() error {
	client, err := docker.NewClient(os.Getenv("DOCKER_HOST"))
	if err != nil {
		return err
	}
	errC := make(chan error, 1)
	statsC := make(chan *docker.Stats)
	go func() {
		errC <- client.Stats(os.Args[1], statsC)
		close(errC)
	}()
	for {
		stats, ok := <-statsC
		if !ok {
			break
		}
		fmt.Printf("%+v\n", stats)
	}
	return <-errC
}
```

fsouza:
As #258 is merged, can you update this pull request? I'm just reviewing it, but you can change the stream calls already.

peter-edge:
I updated but duplicate commits?

peter-edge:
https://github.com/fsouza/go-dockerclient/pull/256

peter-edge:
https://github.com/fsouza/go-dockerclient/pull/262

PR #257, opened 2015-04-19
Makefile

peter-edge:
You probably will reject this one because I'm sure you thought of it, but adding it in case you would find it helpful.

This is a version of what I use in all my repositories, and I think it's a good, universally understood way to show people how to run commands on a repository. The original reason I thought to add this was that I always forget to run `./testing/bin/fmtpolice` when I submit a PR, and then there are usually Travis failures. With this, `fmtpolice` becomes a known part of the testing process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
   - GOARCH=amd64
   - GOARCH=386
 install:
-  - go get -d ./...
+  - make testdeps
 script:
-  - go test ./...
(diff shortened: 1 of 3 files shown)


fsouza:
Thank you very much! Despite the fact that I'm not using much of Makefiles for Go code today, it's a tool that people are used to, so this change would definitely benefit new contributors, specially in the fmtpolice case, as you described.

I'm merging it now :)

PR #258, opened 2015-04-19
update Client#do, Client#stream, Client#hijack to take a struct for optional parameters

peter-edge:
When developing in this repository, I am constantly looking back at client.go to figure out which parameters are what for do, stream, and hijack. Most functions are just passing default values for these anyways, so this PR makes it easier to determine what a call to do, stream, or hijack is doing. This also fixes potential easy such as ordering problems for the same parameter types - stdout and stderr were reversed in stream and hijack, for example.

### DIFF
--- a/client.go
+++ b/client.go
@@ -252,7 +252,7 @@ func (c *Client) checkAPIVersion() error {
 // See http://goo.gl/stJENm for more details.
 func (c *Client) Ping() error {
 	path := "/_ping"
-	body, status, err := c.do("GET", path, nil, false)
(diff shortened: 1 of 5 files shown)


peter-edge:
If https://github.com/fsouza/go-dockerclient/pull/256 is accepted, it will have to be merged with this branch. Personally I'd merge this one first, then I will fix 256 right away,

fsouza:
Nice change, thank you! I went crazy with this sometimes as well, but never really took the opportunity to just refactor the code.

Issue #251, opened 2015-04-10
Can't build image using Dockerfile


nlamirault:
hi
hi try to build an image from a Dockerfile : 

```go
var buf bytes.Buffer
opts := docker.BuildImageOptions{
	Name:           "clyde/" + name,
	Dockerfile:     "/tmp/foo/Dockerfile",
	SuppressOutput: true,
	OutputStream:   &buf,
}
err := client.BuildImage(opts)
if err != nil {
	return err
}
log.Debugf("Docker image: %s", buf.String())
```

I've got an error : 

```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x5cd022]

goroutine 32 [running]:
runtime.panic(0x7fc800, 0xab3413)
        /usr/local/go/src/pkg/runtime/panic.c:279 +0xf5
github.com/fsouza/go-dockerclient.(*Client).stream(0xc2080ee100, 0x854790, 0x4, 0xc2080d8cb0, 0x6f, 0xc2080d0001, 0xc2080d71a0, 0x7f82e61c34b0, 0xc2080d74a0, 0x7f82e61c1aa8, ...)
        /home/vagrant/Apps/clyde/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/client.go:389 +0x10c2
github.com/fsouza/go-dockerclient.(*Client).BuildImage(0xc2080ee100, 0xc2080e2800, 0x9, 0x896ad0, 0x13, 0x100, 0x0, 0x0, 0x7f82e61c1aa8, 0xc2080d8930, ...)
        /home/vagrant/Apps/clyde/Godeps/_workspace/src/github.com/fsouza/go-dockerclient/image.go:386 +0x47f
```

Any idea ?
Thanks.

fsouza:
Hmm, that's interesting. This should not be happening, do you get any error or just the panic? I will try to run your sample locally.

Thanks for reporting!

fsouza:
@nlamirault is your Docker server bound to a TCP port or Unix socket?

Thanks!

nlamirault:
I've got only this : 

```bash
ls /etc/docker/
certs.d
```

```bash

$ cat /usr/lib/systemd/system/docker.socket
[Unit]
Description=Docker Socket for the API
PartOf=docker.service

[Socket]
ListenStream=/var/run/docker.sock
SocketMode=0660
SocketUser=root
SocketGroup=docker

[Install]
[...]

PR #259, opened 2015-04-19
fix nil panic if no response or response body in Client#do or Client#stream

peter-edge:
I believe this at least temporarily solves https://github.com/fsouza/go-dockerclient/issues/251

### DIFF
--- a/client.go
+++ b/client.go
@@ -34,6 +34,9 @@ var (
 	// ErrConnectionRefused is returned when the client cannot connect to the given endpoint.
 	ErrConnectionRefused = errors.New("cannot connect to Docker endpoint")
 
+	// ErrNoResponse is returned when there is no http response.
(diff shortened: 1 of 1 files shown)


fsouza:
@peter-edge are you able to reproduce this locally?

The point is: this should not be happening, as ClientConn.Do/HTTPClient.Do should not return nil, nil.

peter-edge:
I am not, I just saw multiple issues that referenced this. I totally understand the philosophy that if you should understand what is actually going on, and respect if this should not be merged because of that. My instinct was that this is a quick fix for now, which is not necessarily good. Your call :)

fsouza:
@peter-edge that's fine, thanks for working on this. Maybe I'm losing some edge case where Do does return ``nil, nil``, so your code would definitely make sense. I will do some stress testing on it, thank you!

PR #262, opened 2015-04-23
Container Stats #2

peter-edge:


fsouza:
Hi @peter-edge, thanks for working on this. I will review these changes later today or tomorrow morning.

AndrewGuenther:
The only way to stop getting stats in this implementation (before the container is removed) would be to close the channel you passed in, causing a panic in the goroutine.

It would make more sense (to me at least) for this function to return a channel to the caller and accept a `done` channel as a parameter. This function can then exit gracefully once a message has been posted to that `done` channel.

peter-edge:
go-dockerclient does not have cancellation implemented for any of the streaming methods yet. This is desirable, and we should do it. There's a PR out there for the start of it https://github.com/fsouza/go-dockerclient/pull/208

In terms of returning the channel vs not, I'm not sure. This function more closely matches other functions on Client, namely AddEventListener. I wanted to stay consistent.

raychaser:
It would be great if this could get merged. I could really use it ;) - been playing with it today and it appears to work ok.

container.go
```diff
@@ -542,6 +542,128 @@ func (c *Client) TopContainer(id string, psArgs string) (TopResult, error) {
...
+
+// Stats sends container statistics for the given container to the given channel.
+//
+// This function is blocking, similar to a streaming call for logs, and should be run
+// on a separate goroutine from the caller. Note that this function will block until
+// the given container is removed, not just exited. When finished, this function
+// will close the given channel.
+//
+// See http://goo.gl/DFMiYD for more details.
+func (c *Client) Stats(id string, statsC chan<- *Stats) (retErr error) {
```
fsouza: I would prefer to use a struct with options here, just in case Docker extends the API, so we can add new parameters to the method without breaking existing code. Something like:

```
type StatusOptions struct {
	Container string
	Stats     chan<- *Stats
}
```

At the moment, you won't use to build any querystring, but in the future we might need it.

fsouza:
I'm just worried about the cancellation, but I'm getting back to #208 as well.

@raychaser can you give more details of your use case, just to ensure that we can live without cancellation for this moment? Otherwise, we can focus on getting #208 (or something else) merged, and then get back to this PR.

@peter-edge thanks for contributing, and sorry about the delay!

raychaser:
@fsouza i am trying to put together a simple way to transport logs, stats, and daemon events to a centralized logging system. full disclosure, we happen to have a hosted one at http://www.sumologic.com where i work. i am using Logspout for the logs, and was hoping to prototype an extension that also pulls the stats. so, i am in it to get all the stats for the life of the container, and therefore can live without the cancellation, i think. i am sure it is useful in general, but for me right now, i'd rather have the raw capability, and not wait for the cancellation. that's just me tho.

peter-edge:
I was actually working on a fixing up of https://github.com/fsouza/go-dockerclient/pull/208 after this, and working to understand net.Context. My hope was to get this in first, and then we can finish up 208. What do you think?

fsouza:
@raychaser thanks for presenting your point! I think it's fine to merge it then.

@peter-edge nice to know! I was going to give it some attention as well. Let's merge this one so. Can you make that struct change I've commented inline? Thanks for contributing.

peter-edge:
Updated. I named the container id field "ID" to more closely match the rest of the API.

fsouza:
Thank you! Merging it now.

Issue #266, opened 2015-04-30
no port mapping host ip


tcurdt:
I am currently tracking down a missing ip on the port binding. Looking at the inspect output I see

                    "HostIp": "0.0.0.0",
                    "HostPort": "6667"

while the code seems to expect `HostIP`

  https://github.com/fsouza/go-dockerclient/blob/master/container.go#L116

s/HostIP/HostIp/g ?

tcurdt:
https://github.com/docker/docker/blob/master/docs/sources/reference/api/docker_remote_api_v1.16.md
https://github.com/docker/docker/blob/master/daemon/container.go#L643

yup!

fsouza:
Closing, see #267.

PR #267, opened 2015-04-30
fixes empty port bindings hostip

tcurdt:
Test are passing - just your formatting police doesn't seem to like the change.
But that's how docker uses it - so I suggest to change the formatting check instead.

### DIFF
--- a/container.go
+++ b/container.go
@@ -113,7 +113,7 @@ func (s *State) String() string {
 // PortBinding represents the host/container port mapping as returned in the
 // `docker inspect` json
 type PortBinding struct {
-	HostIp   string `json:"HostIp,omitempty" yaml:"HostIp,omitempty"`
+	HostIP   string `json:"HostIP,omitempty" yaml:"HostIP,omitempty"`
 	HostPort string `json:"HostPort,omitempty" yaml:"HostPort,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


tcurdt:
fixes https://github.com/fsouza/go-dockerclient/issues/266

fsouza:
Hey @tcurdt, thanks for contributing!

Please do not change the name of the attribute, just the struct tag. Changing the name of the attribute won't complain with the formatting police and will also break existing code.

Just this fix and we're good to go. Thanks again!

fsouza:
BTW, as far as I can tell, Go does handle that (I will test it on master, but at least on Go 1.4 I'm sure it works). Which Go version you're running? I'm OK with fixing the struct tag, but maybe this won't fix the issue you're facing, have you tested it?

fsouza:
Yeah, just tested and this code [1] always print "0.0.0.0 8080", both at Go 1.4 and Go tip. Can you confirm that this patch fixes the empty HostIP issue you're fighting?

[1] https://play.golang.org/p/pBxXqSIAR0

tcurdt:
Changed as requested.
Just fixing the json/yaml mapping would mean a broken window though.

Hm. So casing doesn't actually matter? Urgh!
Then I am not quite sure what's happening and need to dig deeper.
I just found the discrepancy with docker and thought it was the problem.

In [my code](https://github.com/tcurdt/docker-upstream/blob/master/src/upstream/upstream.go#L115) I am getting a port but no IP. So far I cannot see a difference between the two except that `docker inspect` shows `HostIp` instead of `HostIP`.

I am on go 1.4.2.

tcurdt:
OK. Found out what's going on.

The host IP actually really was empty in `HostConfig.PortBindings` (which can be solved by specifying it explicitly on the `docker run` command line). The better way to access the ip/ports would probably be `NetworkSettings.Ports`.

tcurdt:
So I guess I'll just close the PR :)

PR #269, opened 2015-05-05
Allow DockerServer to force remove a container

KAllan357:
Hi @fsouza,

Thanks for this wonderful library. I have been writing some functionality and I've been really digging the testing package.

I noticed a problem though - I wanted to delete a container and when I set the option `force: true`, it still wouldn't delete. I tracked the problem down to the testing server not honoring the query param of `force=1`. 

Here's a PR that fixes the issue!


fsouza:
Nice catch, thank you!

PR #270, opened 2015-05-06
Add /container/ID/logs endpoint to test server

KAllan357:
Hi @fsouza,

I'd like some feedback on this one:

I wanted to test the logs endpoint on the test docker server, to ensure my code was operating correctly using `client.Logs(opts)`.

If I'm understanding correctly, the test server is writing out data using the code from [writer.go](https://github.com/fsouza/go-dockerclient/blob/master/testing/writer.go). It looks like that in order to read what the writer has written, I'd want to use the `stdCopy` function in stdcopy.go. Unfortunately, the `stdCopy` function is private in this repo.

I'm guessing it might be copied from docker themselves (https://github.com/docker/docker/blob/master/pkg/stdcopy/stdcopy.go) where its public. Perhaps you'd be okay with making `stdCopy` a public function in this repo so the test can strip off the extra docker bytes?

Anyway, besides any feedback you might have, I think there is room for a little more testing here.

### DIFF
--- a/client.go
+++ b/client.go
@@ -438,7 +438,7 @@ func (c *Client) stream(method, path string, streamOptions streamOptions) error 
 		if streamOptions.setRawTerminal {
 			_, err = io.Copy(streamOptions.stdout, resp.Body)
 		} else {
-			_, err = stdCopy(streamOptions.stdout, streamOptions.stderr, resp.Body)
+			_, err = StdCopy(streamOptions.stdout, streamOptions.stderr, resp.Body)
 		}
 		return err
 	}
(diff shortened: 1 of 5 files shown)


KAllan357:
Sorry about that - I fixed the missing references to the renamed function. I'm sorta assuming this code is just a copy of what Docker has in their lib, so I just copied the GoDoc for it as well.

KAllan357:
I tried some hackery with using CustomHandler like so:

```
fakeServer.CustomHandler("/containers/"+dockerID+"/logs", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	w.Header().Add("Content-Type", "application/vnd.docker.raw-stream")
	fmt.Fprintln(w, "log output")
	fmt.Fprintln(w, "more log output")
}))
```

but my calling code gets an error on `Unrecognized input header` because my options aren't setting `RawTerminal` to true.

fsouza:
Hi @KAllan357, in order to write a CustomHandler, you'd need to write the stream header before writing the content, just like stdWriter does.

fsouza:
For testing, you do not need the StdCopy function. You can use an actual client with the testing server and check whether the Logs function work as expected. I'd also like to be able to customize the logging from a container, without having to provide a custom handler. Something like:

```
func (*DockerServer) PrepareLogs(containerID string, stdout, stderr io.Reader)
```

This method would read the contents of stdout and stderr and store it for future Logs calls.

fsouza:
@KAllan357 are you going to send another pull request with this change? Thank you!

PR #271, opened 2015-05-07
Add blkio_stats to Stats struct

bobrik:
I only have the following stats in 1.6.0:

```
~ λ curl -sN http://127.0.0.1:2375/containers/e2ae202a2a41/stats | head -n1 | jq .blkio_stats
```

```json
{
  "io_service_bytes_recursive": [
    {
      "major": 8,
      "minor": 0,
      "op": "Read",
      "value": 428795731968
    },
    {
      "major": 8,
      "minor": 0,
      "op": "Write",
      "value": 388177920
    },
    {
      "major": 8,
      "minor": 0,
      "op": "Sync",
      "value": 387575808
    },
    {
      "major": 8,
      "minor": 0,
      "op": "Async",
      "value": 428796334080
    },
    {
      "major": 8,
      "minor": 0,
      "op": "Total",
      "value": 429183909888
    }
  ],
  "io_serviced_recursive": [
    {
      "major": 8,
      "minor": 0,
      "op": "Read",
      "value": 25994442
    },
    {
      "major": 8,
      "minor": 0,
      "op": "Write",
      "value": 1734
    },
    {
      "major": 8,
      "minor": 0,
      "op": "Sync",
      "value": 1666
    },
    {
      "major": 8,
      "minor": 0,
      "op": "Async",
      "value": 25994510
    },
    {
      "major": 8,
      "minor": 0,
      "op": "Total",
      "value": 25996176
    }
  ],
  "io_queue_recursive": [],
  "io_service_time_recursive": [],
  "io_wait_time_recursive": [],
  "io_merged_recursive": [],
  "io_time_recursive": [],
  "sectors_recursive": []
}
```

All entries seem to have the [same structure](https://github.com/docker/libcontainer/blob/59eb58b640c4685265e9ff08734a3e8848a8232c/cgroups/fs/blkio.go#L166-L210).

I'm going to use that in [bobrik/collectd-docker](https://github.com/bobrik/collectd-docker).

### DIFF
--- a/container.go
+++ b/container.go
@@ -596,8 +596,16 @@ type Stats struct {
 		Failcnt  int64 `json:"failcnt,omitempty" yaml:"failcnt,omitempty"`
 		Limit    int64 `json:"limit,omitempty" yaml:"limit,omitempty"`
 	} `json:"memory_stats,omitempty" yaml:"memory_stats,omitempty"`
-	// TODO(pedge): this is in the docker docs, but no data
-	//BlkioStats string `json:"blkio_stats,omitempty" yaml:"blkio_stats,omitempty"`
+	BlkioStats struct {
+		IOServiceBytesRecursive []BlkioStatsEntry `json:"io_service_bytes_recursive,omitempty" yaml:"io_service_bytes_recursive,omitempty"`
+		IOServicedRecursive     []BlkioStatsEntry `json:"io_serviced_recursive,omitempty" yaml:"io_serviced_recursive,omitempty"`
+		IOQueueRecursive        []BlkioStatsEntry `json:"io_queue_recursive,omitempty" yaml:"io_queue_recursive,omitempty"`
+		IOServiceTimeRecursive  []BlkioStatsEntry `json:"io_service_time_recursive,omitempty" yaml:"io_service_time_recursive,omitempty"`
+		IOWaitTimeRecursive     []BlkioStatsEntry `json:"io_wait_time_recursive,omitempty" yaml:"io_wait_time_recursive,omitempty"`
+		IOMergedRecursive       []BlkioStatsEntry `json:"io_merged_recursive,omitempty" yaml:"io_merged_recursive,omitempty"`
+		IOTimeRecursive         []BlkioStatsEntry `json:"io_time_recursive,omitempty" yaml:"io_time_recursive,omitempty"`
+		SectorsRecursive        []BlkioStatsEntry `json:"sectors_recursive,omitempty" yaml:"sectors_recursive,omitempty"`
+	} `json:"blkio_stats,omitempty" yaml:"blkio_stats,omitempty"`
 	CPUStats struct {
 		CPUUsage struct {
 			PercpuUsage       []int64 `json:"percpu_usage,omitempty" yaml:"percpu_usage,omitempty"`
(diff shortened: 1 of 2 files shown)


peter-edge:
Can you add the data above to to the test at https://github.com/fsouza/go-dockerclient/blob/0ba7f0a32981ca4c1043c245017eb099708695ef/container_test.go#L1561 and https://github.com/fsouza/go-dockerclient/blob/0ba7f0a32981ca4c1043c245017eb099708695ef/container_test.go#L1625? Otherwise LGTM @fsouza

bobrik:
Updated the test. Did I get it right that test only checks types of struct? Commenting out random parts of the `Stats` struct doesn't make the test fail.

peter-edge:
Yep, you got it right. It's a really bad test, but wanted something basic in there for now. LGTM

peter-edge:
Hey while you're in there, there's two other fields I had commented out, https://github.com/fsouza/go-dockerclient/blob/0ba7f0a32981ca4c1043c245017eb099708695ef/container.go#L581 and https://github.com/fsouza/go-dockerclient/blob/0ba7f0a32981ca4c1043c245017eb099708695ef/container.go#L610, there's probably docker equivalents in the docker code (probably in the same area that your found blkio stats), if you had a few minutes do you think you could add those as well? No worries if not.

bobrik:
What exactly was wrong with `hierarchical_memory_limit `? On my system it's under 2^63 and it seems to work just fine. It's `uint64` in [docker codebase](https://github.com/docker/libcontainer/blob/59eb58b640c4685265e9ff08734a3e8848a8232c/cgroups/stats.go) so json decoding could easily fail.

Throttling data is empty (zeroes) for me as well, [struct](https://github.com/docker/libcontainer/blob/59eb58b640c4685265e9ff08734a3e8848a8232c/cgroups/stats.go#L3-L10) has `uint64` as well.

Maybe it's a good idea to replace all `int64` with `uint64`. Anyway, it deserves another PR.

peter-edge:
The int64/uint64 thing was to match the rest of the repo. Docker actually
uses in64 in a lot of places too. This I didn't like, but I wanted to be
consistent.

I don't remember what was wrong - if you comment it out and do a quick test
with it, it froze for me.

bobrik:
```
--- FAIL: TestStats (0.00s)
	container_test.go:1765: json: cannot unmarshal number 9223372036954775807 into Go value of type int64
```

This only happens with values bigger than int64 can hold, with smaller values all tests pass.

peter-edge:
Have you tried actually using it with a running docker container? And then
killing that docker container? I think this was my issue.

More importantly, we should have an actual integration test for this, but
ya, separate PR.

bobrik:
Just tried with actual docker daemon — it works just fine, stopping and starting container didn't break anything.

peter-edge:
Well there you go :) feel free to update.

Also note I'm not a maintainer of this repo, that's @fsouza's department - I just happened to implement this feature.

fsouza:
Ops, sorry for the poor review in the first place, I will take this pull request, fix both @peter-edge's TODOs  and change everything to uint64. This may break some existing code, but will make it actually compatible with Docker.

peter-edge:
@fsouza I actually did make them int64 to match the rest of the repository, I'm not sure if it actually does match. I would have preferred uint64 everywhere for these things so if we're doing this, we should make sure 1. we're matching with github.com/docker/docker for any changes 2. changing this universally in the repository.

fsouza:
As far as I remember, there was a match, Docker used to have int64 for the amount of memory or swap, among other things. I don't think it has changed, because Docker used to use -1 to signal disabled swap, for example, and I hope they didn't break their API (usually they do).

I will also have a look at these other things, but this should not be the case to check just whether we use int64 or int64, but actually understanding the value. If I had the power to choose, I'd use uint64 in anything that would probably become big enough and never less than zero, but we can not control this :-(

peter-edge:
+1

I just want to make sure we're not breaking anything. Someone (probably
someone on this thread? :) ) should spend an hour or two and do an audit on
this.

PR #273, opened 2015-05-08
Have .dockerignore support Dockerfile/.dockerignore

ybogdanov:
### Problem
Sometimes you have to add `Dockerfile` itself to `.dockerignore`, in cases like this https://github.com/docker/docker/issues/8330

For that purpose, the following changes were added to Docker: https://github.com/docker/docker/commit/6d801a3caa54ad7ef574bc426aa1ffc412c5af82

It became critical for me as well, so I decided to implement it also for `go-dockerclient`.

### Solution
Docker client will read `.dockerignore` and whilelist `Dockerfile` or `.dockerignore` to make sure they will be transferred to a server. Server is reading `.dockerignore` again, and removing those files after capturing them in memory https://github.com/docker/docker/blob/596e91638cbe8b583dff4fdfc5a52e9d3840b20f/builder/evaluator.go#L240

### DIFF
--- a/tar.go
+++ b/tar.go
@@ -103,16 +103,7 @@ func parseDockerignore(root string) ([]string, error) {
 	if err != nil && !os.IsNotExist(err) {
 		return excludes, fmt.Errorf("error reading .dockerignore: '%s'", err)
 	}
-	for _, pattern := range strings.Split(string(ignore), "\n") {
-		matches, err := filepath.Match(pattern, "Dockerfile")
-		if err != nil {
-			return excludes, fmt.Errorf("bad .dockerignore pattern: '%s', error: %s", pattern, err)
-		}
-		if matches {
-			return excludes, fmt.Errorf("dockerfile was excluded by .dockerignore pattern '%s'", pattern)
-		}
-		excludes = append(excludes, pattern)
-	}
+	excludes = strings.Split(string(ignore), "\n")
 
 	return excludes, nil
 }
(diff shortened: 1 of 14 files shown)


tar.go
```diff
@@ -17,17 +17,36 @@ import (
...
+	// If .dockerignore mentions .dockerignore or the Dockerfile
+	// then make sure we send both files over to the daemon
+	// because Dockerfile is, obviously, needed no matter what, and
+	// .dockerignore is needed to know if either one needs to be
+	// removed.  The deamon will remove them for us, if needed, after it
+	// parses the Dockerfile.
+	//
+	// https://github.com/docker/docker/issues/8330
+	//
+	if keepThem, _ := fileutils.Matches(".dockerignore", excludes); keepThem {
```
peter-edge: check error, do not ignore, this will fail errcheck

tar.go
```diff
@@ -17,17 +17,36 @@ import (
...
+		includes = append(includes, ".dockerignore")
+	}
+	if keepThem, _ := fileutils.Matches(dockerfileName, excludes); keepThem && dockerfileName != "" {
```
peter-edge: same

tar.go
```diff
@@ -17,17 +17,36 @@ import (
 	"github.com/docker/docker/pkg/fileutils"
 )
 
-func createTarStream(srcPath string) (io.ReadCloser, error) {
+func createTarStream(srcPath, dockerfileName string) (io.ReadCloser, error) {
```
peter-edge: nit: probably would change to dockerfilePath

fsouza:
Hi @ybogdanov, thanks for contributing!

@peter-edge thanks for reviewing it, it's good to merge! :-)

PR #274, opened 2015-05-10
Update makefile to use lint, vet, errcheck, gofmt, and remove testing/bin/fmtpolice

peter-edge:
This continues the movement to the Makefile standard for go-dockerclient - after this PR, I can immediately see all options that are done for testing in the Makefile, without having to know or read testing/bin/fmtpolice. This also is a simpler version of fmtpolice, other than perhaps gofmt :)

A few errors were caught by vet and errcheck. The errcheck errors are too long right now, so I'll work on cleaning those up (mostly not checking for error on close of an object). But this has already been useful.


peter-edge:
Note that after this, I have a branch chained on top of this branch that handles or explicitly ignores every error return, so that this passes errcheck. I actually found a few testing errors by doing this, and I think that existing functionality stays the same. There is an extra check in hijack, but overall every error should be handled. The ignored errors are generally Write calls in tests.

https://github.com/peter-edge/go-dockerclient/compare/makefile...peter-edge:makefile2

peter-edge:
Note that on that branch, I also updated the README to reflect new, stricter testing procedures https://github.com/peter-edge/go-dockerclient/blob/makefile2/README.markdown

fsouza:
Hi @peter-edge, thanks for working on this!

Please remove errcheck, it doesn't catch anything useful, and in order to make it happy we need to make the code uglier and more complex with no change in semantics (we still ignore the errors in Close calls, but now there's a function for that).

peter-edge:
Hey,

I'll remove it, but note that it actually did catch some useful things in makefile2 - I can just copy over those things if you want?

peter-edge:
Also note that we are not ignoring the errors in Close calls, I actually do check them now

peter-edge:
Actually ya, let me know if you want the closes checked, I think they should be. We'll disagree on this, but it's your repository so your decision :) I think all errors should be checked, even if it's uglier. It's more strict but it causes things to be missed if you don't. I found a few testing bugs while going through makefile2.

fsouza:
Sorry, I messed up some changes in test (like ``_ = l.Close()`` and ``_, _ = w.Write``) with the deferCheckError function.

It depends on the Close failure. Most of the times that a closing fail, I can't do much things (besides logging).

I understand that logging is important sometimes, but for the cases the branch is handling (clientconn and response body), what's the user of the library supposed to do? Everything went fine, the request has been successfully handled by Docker and the client failed to close the connection, should the user code consider that the operation has failed?

peter-edge:
Ya, that's a good question. Probably not.

So I can remove the Close checks where appropriate, and put a comment that we're not handling it explicity because we do not consider it a user error. I still am strongly in favor of errcheck but can remove if you want :)

fsouza:
That's fine, but we can go with no comments, just the code as it is today. Please also remove errcheck, we can live without it.

Thanks!

peter-edge:
OK, removed. I'll do a separate PR for further cleanups regarding this, I'll cherry pick a few things from makefile2

peter-edge:
Fixed

peter-edge:
What the hell, the rebase keeps on getting messed up - I can do a patch and make this one commit if you want.

fsouza:
I think I messed upstream with a no ``--no-ff`` merge, sorry :-(

peter-edge:
I'll do a patch, I don't want to mess up the commit history.

fsouza:
Thank you!

PR #275, opened 2015-05-10
Use shields.io for badges when possible

peter-edge:
You can totally just reject this one :) but while I was doing updates for ya...

http://shields.io/
https://github.com/badges/shields

There isn't a drone one yet but I think they are working on it, luckily the style actually matches.


fsouza:
Ah, that's fair. Thank you!

Issue #260, opened 2015-04-19
Remove dependency on github.com/docker/docker


peter-edge:
One of the biggest pains for me with go-dockerclient is the dependency on github.com/docker/docker. The only thing that is used is the tar functionality in https://github.com/docker/docker/tree/master/pkg/archive, and a couple of file utilities (Matches, Rel) that can be easily redone.

The problem is, docker's tar implementation is probably the best I know of in the golang world. So we don't want to lose that. Also, the archive package depends on a lot of sub-packages:

github.com/docker/docker/pkg/archive
github.com/docker/docker/pkg/fileutils
github.com/docker/docker/pkg/ioutils
github.com/docker/docker/pkg/pools
github.com/docker/docker/pkg/promise
github.com/docker/docker/pkg/system
github.com/docker/docker/vendor/src/code.google.com/p/go/src/pkg/archive/tar

However, having to download docker/docker each time I want to make a new container is a real big pain.

Since this repo already depends on the docker license, could we literally vendor out that code and put it in a sub-package here? I don't think docker will ever split up their repo, so that might be the only way. The unfortunate part is that we lose changes upstream to pkg/, but I am almost willing to have that happen. We could have a script in here that basically does:

```
go get -u github.com/docker/docker
cp -R $GOPATH/github.com/docker/docker/pkg/* vendor/docker/pkg/
cp -R $GOPATH/github.com/docker/docker/vendor/src/code.google.com/p/go/src/pkg/archive/tar/ ... # you get the idea
```

Just run it every so often to selectively get updates.

Maybe there is a better way to do this, but whenever I'm running my code inside a Docker container, and my code relies on go-dockerclient and I need to get all the dependencies, there's a huge wait to get github.com/docker/docker. It's just no fun, and all just for some tar functionality.

fgrehm:
Not sure if you came across that yet but here's some background on it: GH-178

peter-edge:
Ah shoot.

The primary reason this is such a pain is when you're building new Docker images with this, library, ironically. Would you be open to a script that just did a selective pre-commit copy, like above?

fsouza:
@peter-edge how often do you need to build a new Docker image? What's your processing of doing it?

Sorry for the delay!

archisgore:
+1 on this. This adds resource cost to build, download and even godep save to be pulling on ALL of docker and building it every time.

peter-edge:
[...]

PR #277, opened 2015-05-10
Use https://github.com/kardianos/vendor for external packages

peter-edge:
I just found this, it's an extremely clean solution to https://github.com/fsouza/go-dockerclient/issues/260.

Not only does it solve that issue, but it makes go-dockerclient:
* Completely self-contained. No dependencies are needed to use go-dockerclient, and I updated the makefile accordingly.
*Revison-locked. All external dependencies are externally locked, so go-dockerclient is not susceptible to external bugs.

Updating is extremely simple, I added `make vendor` to the Makefile. That's all it is.

This adds a ton of LOC, and since you're the maintainer, I don't want to muck up the contributors, so if you want to just do this instead, just do this:

```bash
go get -v github.com/kardianos/vendor
vendor init
vendor add -status external
```

And then I can update the Makefile in a separate PR, or just copy what is there right now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,5 @@ go:
 env:
   - GOARCH=amd64
   - GOARCH=386
-install:
-  - make testdeps
 script:
   - make test
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ all: test
 
 vendor:
 	go get -v github.com/kardianos/vendor
+	vendor init
 	vendor add -status external
 	vendor update -status external
 
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ all: test
 
 vendor:
 	go get -v github.com/kardianos/vendor
-	vendor init
 	vendor add -status external
 	vendor update -status external
 
--- a/Makefile
+++ b/Makefile
@@ -1,26 +1,16 @@
 .PHONY: \
 	all \
-	deps \
-	updatedeps \
-	testdeps \
-	updatetestdeps \
+	vendor \
 	cov \
 	test \
 	clean
 
 all: test
 
-deps:
-	go get -d -v ./...
-
-updatedeps:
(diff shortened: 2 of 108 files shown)


peter-edge:
Note that the actual diff in terms of go-dockerclient code is only this (and this was automatically done by the vendor command):

```
$ git diff
diff --git a/build_test.go b/build_test.go
index ecd9885..3e5232d 100644
--- a/build_test.go
+++ b/build_test.go
@@ -9,7 +9,7 @@ import (
        "reflect"
        "testing"

-       "github.com/docker/docker/pkg/archive"
+       "github.com/fsouza/go-dockerclient/internal/github.com/docker/docker/pkg/archive"
 )

 func TestBuildImageMultipleContextsError(t *testing.T) {
diff --git a/tar.go b/tar.go
index aa56be2..eda2993 100644
--- a/tar.go
+++ b/tar.go
@@ -13,8 +13,8 @@ import (
        "path/filepath"
        "strings"

-       "github.com/docker/docker/pkg/archive"
-       "github.com/docker/docker/pkg/fileutils"
+       "github.com/fsouza/go-dockerclient/internal/github.com/docker/docker/pkg/archive"
+       "github.com/fsouza/go-dockerclient/internal/github.com/docker/docker/pkg/fileutils"
 )

 func createTarStream(srcPath string) (io.ReadCloser, error) {
diff --git a/testing/server.go b/testing/server.go
index 1c449e5..c68a06e 100644
--- a/testing/server.go
+++ b/testing/server.go
@@ -22,7 +22,7 @@ import (
        "time"

        "github.com/fsouza/go-dockerclient"
-       "github.com/gorilla/mux"
+       "github.com/fsouza/go-dockerclient/internal/github.com/gorilla/mux"
 )

 var nameRegexp = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9_.-]+$`)
```

PR #278, opened 2015-05-10
Use https://github.com/mjibson/party to vendor external packages

peter-edge:
Better version of https://github.com/fsouza/go-dockerclient/pull/277. This does not download test dependencies for external packages, and since the directory is prefaced with _, `go test` does not pick up those tests for testing. Additionally, using `-c` for create is allowed even if `_third_party` is already created, so the entire command for vendoring can easily be put in the Makefile.

I just found this, it's an extremely clean solution to https://github.com/fsouza/go-dockerclient/issues/260.

Not only does it solve that issue, but it makes go-dockerclient:

* Completely self-contained. No dependencies are needed to use go-dockerclient, and I updated the makefile accordingly.
*Revison-locked. All external dependencies are externally locked, so go-dockerclient is not susceptible to external bugs.

Updating is extremely simple, I added `make vendor` to the Makefile. That's all it is.

This adds a ton of LOC, and since you're the maintainer, I don't want to muck up the contributors, so if you want to just do this instead, just do the commands on `make vendor` and copy the changes I made to `.travis.yml, Makefile, testing/bin/fmtpolice`. Very easy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,5 @@ go:
 env:
   - GOARCH=amd64
   - GOARCH=386
-install:
-  - make testdeps
 script:
   - make test
--- a/Makefile
+++ b/Makefile
@@ -1,33 +1,22 @@
 .PHONY: \
 	all \
-	deps \
-	updatedeps \
-	testdeps \
-	updatetestdeps \
+	vendor \
 	cov \
 	test \
 	clean
 
 all: test
 
-deps:
-	go get -d -v ./...
+vendor:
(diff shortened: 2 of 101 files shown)


fsouza:
That's indeed a clean solution, I dislike vendoring in libraries, but since it's a recurrent topic, I will take this pull request. It's fine to have you as the committer of these changes.

Thank you very much!

peter-edge:
The beautiful thing is it doesn't even require the entire github repository for each package - this just takes the individual packages. Such a win! :) Thanks :)

PR #279, opened 2015-05-10
lint, fmt, vet, etc

peter-edge:
Fix of previous PR. Also brought over some stuff from makefile2 that I thought should be added - not checking writes/closes, but errcheck did catch a few things :)


peter-edge:
All ready/merged

fsouza:
Thanks!

container_test.go
```diff
@@ -1091,10 +1093,7 @@ func TestAttachToContainerRawTerminalFalse(t *testing.T) {
 		Stream:       true,
 		RawTerminal:  false,
 	}
-	err := client.AttachToContainer(opts)
-	if err != nil {
-		t.Fatal(err)
-	}
+	client.AttachToContainer(opts)
```
fsouza: This was backwards? :-)

peter-edge: Ah no it wasn't, this was a new error because now we check the io.Copy, I missed that this was not backwards...working on it now, my mistake.

peter-edge: Ah nevermind, it's on the stdCopy, because now we actually catch that error (correctly)...the fake data in the response can't be parsed. I think this is fine relative to what this test is testing - this was broken before anyways I think.

fsouza: I noticed while editing it, just fixed the test.

Also changed the code to use Docker's StdCopy (it's now exported there).

PR #280, opened 2015-05-10
Vendor fix!

peter-edge:
If other libraries do their own vendoring with Godeps (which I do, for example, and I'm sure a lot of others do), having the vendored files in _third_party/ will make them not be picked up and break Godeps. Moved to vendor/ and updated the tests.

This should be merged ASAP.


fsouza:
Thanks!

PR #286, opened 2015-05-16
Add resource isolation parameters to HostConfig.

### LINKED ISSUE #285: Resource requirement fields in Config should be in HostConfig
Things like `CpuShares` and `MemoryLimit` don't exist in Docker's `Config`, rather they are in `HostConfig` as seen in their [documentation](https://docs.docker.com/reference/api/docker_remote_api_v1.18/#create-a-container). We should switch move those fields to `HostConfig` where we can make use of them.

Will take a crack at a PR.

GoogleCloudPlatform/kubernetes#8358

/cc @dchen1107

vmarmol:
This was moved in the v1.18 API. Current implementation is backwards and
forwards compatible, having the fields for older and newer versions of
the API.

Verified fix on new Docker API.

Fixes #285.


vmarmol:
Ping @fsouza :)

fsouza:
Thank you! Sorry about the delay, the weekend was longer than expected x)

vmarmol:
No worries, thanks for the merge! :)

dchen1107:
Thanks for the fix.

PR #291, opened 2015-05-22
add client method to check credentials

daniel-garcia:
This PR adds a method for checking credentials.

https://docs.docker.com/reference/api/docker_remote_api_v1.18/#check-auth-configuration

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -7,6 +7,7 @@ package docker
 import (
 	"encoding/base64"
 	"encoding/json"
+	"fmt"
 	"io"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


daniel-garcia:
ping @fsouza , any feedback on this PR?

fsouza:
Hi @daniel-garcia, thanks for contributing and sorry about the delay. I was wondering whether we should read the body of the response and include it in the error message. According to the documentation, Docker will simply return 500 on failures, I think the reason of the failure will come in the body.

daniel-garcia:
@fsouza, reading the body already happens on errors as part of Client.do()
https://github.com/fsouza/go-dockerclient/blob/master/client.go#L344

fsouza:
@daniel-garcia what I meant is including the body in the error report, sorry. Something like:

```go
body, statusCode, err := c.do("POST", "/auth", doOptions{data: conf})
if err != nil {
	return err
}
if statusCode > 400 {
	return fmt.Errorf("auth error (%d): %s", statusCode, body)
}
```

daniel-garcia:
@fsouza thanks for clarifying. I've updated the PR as you requested.

fsouza:
Thank you! Sorry about the miscommunication.

PR #292, opened 2015-05-24
Add codeship files

peter-edge:
We're starting the services integration beta on Wednesday so pre-emptive move to get go-dockerclient ready :) @fsouza


fsouza:
Thanks @peter-edge, I'm merging the pull request now.

Is it live already?

PR #293, opened 2015-05-26
Add NewTLSClientInline, NewTLSVersionedClientInline

peter-edge:
This allows certificates and keys to be specified from byte slices instead of local file paths. This is really useful, for example, if go-dockerclient is used within a library that does not necessarily use local file storage - certificates could be stored in a secured database.

I'm not sure on the function name, but this is the best I could come up with. This function really is kind of necessary at least for me :)

This is how it is equivalent: https://golang.org/src/crypto/tls/tls.go?s=4745:4821#L160

### DIFF
--- a/client.go
+++ b/client.go
@@ -153,6 +153,18 @@ func NewTLSClient(endpoint string, cert, key, ca string) (*Client, error) {
 	return client, nil
 }
 
+// NewTLSClientInline returns a Client instance ready for TLS communications with the givens
+// server endpoint, key and certificates (passed inline to the function as opposed to being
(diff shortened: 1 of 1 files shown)


fsouza:
Hi @peter-edge, thanks for contributing! I was thinking about the name, what do you think of "NewTLSFromBytes" instead of "...Inline"?

peter-edge:
I can do that :)

fsouza:
Cool! Please make this change and I will merge the pull request.

Thanks for contributing! :-)

peter-edge:
Done :)

fsouza:
Thank you!

PR #297, opened 2015-06-04
Switch version call to using map[string]interface{} for API 1.19

ibuildthecloud:
Docker 1.7, API 1.19 introduces a new fields for "Experimental":
false/true.  This means the response is not a map[string]string and
unmarshalling will fail.  We switch to map[string]interface{} to be
compatible with mixed types.

The Version() call in misc.go is not impacted by this as it returns an
Env.


ibuildthecloud:
Refer to https://github.com/docker/docker/issues/13730

fsouza:
Thank you!

PR #299, opened 2015-06-05
refactor ParseEndpoint func

dalanlan:
Refactor to gain readability and conciseness.

### DIFF
--- a/client.go
+++ b/client.go
@@ -680,7 +680,6 @@ func parseEndpoint(endpoint string, tls bool) (*url.URL, error) {
 	if tls {
 		u.Scheme = "https"
 	}
-
 	switch u.Scheme {
 	case "unix":
 		return u, nil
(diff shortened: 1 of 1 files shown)


peter-edge:
just a quick thought: as long as you're doing this, this would be much clearer/more golang idiomatic (IMO) as a switch statement:

```go
switch u.scheme {
case "unix"
  return u, nil
case "http":
  fallthrough
case "https":
  fallthrough
case "tcp":
  ...
default:
  return nil, ErrInvalidEndpoint
}
```

what do you think?

peter-edge:
oh, wait, you don't need those fallthroughs, i forgot:

```go
switch u.scheme {
case "unix"
  return u, nil
case "http", "https", "tcp":
  ...
default:
  return nil, ErrInvalidEndpoint
}
```

fsouza:
Hi @dalanlan, thanks for contributing! I liked @peter-edge's idea, what do you think?

dalanlan:
Make sense :) 
i did not think further at the very beginnig and will fix it later.

dalanlan:
Updated. 
@fsouza

fsouza:
@dalanlan thank you!

Issue #300, opened 2015-06-05
streaming calls never return


cjellick:
This is related to, but different from #298.
That one is a feature request for a way to explicitly terminate Logs() (and Stats()) calls.
This is a bug report.

When making a streaming call such as Logs() or Stats(), if I close the writer I passed in for stdout or stderr, Logs() should return with an error, but it continues to block.

This is because this line:
https://github.com/fsouza/go-dockerclient/blob/7b927cb79baf026dd14d1c507033c33e5cf74eed/client.go#L432
blocks forever.

It blocks forever because you apparently cannot close the Body until the entire thing has been read.

PR #301, opened 2015-06-05
Fix for blocked Body.Close()

cjellick:
This change fixes a bug that was preventing client.stream() from returning properly when the supplied writer closed.

The issue is that you cannont call Body.Close() until the entire body is read. Since this
is a streaming request, the end of the body is never reached unless we
close the connection from the client side.

Addresses #300

### DIFF
--- a/client.go
+++ b/client.go
@@ -119,6 +119,7 @@ func (version APIVersion) compare(other APIVersion) int {
 type Client struct {
 	SkipServerVersionCheck bool
 	HTTPClient             *http.Client
+	transport              *http.Transport
 	TLSConfig              *tls.Config
 
 	endpoint            string
@@ -179,8 +180,11 @@ func NewVersionedClient(endpoint string, apiVersionString string) (*Client, erro
 			return nil, err
 		}
 	}
+
+	tr := &http.Transport{}
 	return &Client{
 		HTTPClient:          http.DefaultClient,
+		transport:           tr,
 		endpoint:            endpoint,
 		endpointURL:         u,
 		eventMonitor:        new(eventMonitoringState),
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -429,7 +437,6 @@ func (c *Client) stream(method, path string, streamOptions streamOptions) error
 		}
 		return err
 	}
-	defer resp.Body.Close()
```
cjellick: Needed to move Body.Close() because defers are Last-in, First-out and we want the connection closed before closing the body.

client.go
```diff
@@ -419,9 +424,12 @@ func (c *Client) stream(method, path string, streamOptions streamOptions) error
 		}
 		clientconn := httputil.NewClientConn(dial, nil)
```
dalanlan: It is said that ClientConn is low-level and old(which `NewClientConn` func returns). Do you mind fixing this btw :)
http://golang.org/pkg/net/http/httputil/#ClientConn

client.go
```diff
@@ -119,6 +119,7 @@ func (version APIVersion) compare(other APIVersion) int {
 type Client struct {
 	SkipServerVersionCheck bool
 	HTTPClient             *http.Client
+	HTTPTransport          *http.Transport
```
fsouza: Does HTTPTransport really need to be exported?

cjellick: No, it doesn't. Good catch.

cjellick:
Ping @fsouza 
I updated to make the transport non-exported.

fsouza:
Thank you!

joeshaw:
This PR introduces a new bug, which I've filed at https://github.com/fsouza/go-dockerclient/issues/318

PR #302, opened 2015-06-09
Return raw JSON if asked for

subhraveti:
This is simple but very useful.  Thanks.


fsouza:
Nice change, thank you!

PR #306, opened 2015-06-21
Adding ulimits configuration to host config (added in api v1.18)

liron-l:
https://docs.docker.com/reference/api/docker_remote_api_v1.18/

See:
Ulimits - A list of ulimits to be set in the container, specified as {
"Name": <name>, "Soft": <soft limit>, "Hard": <hard limit> }, for
example: Ulimits: { "Name": "nofile", "Soft": 1024, "Hard", 2048 }}


fsouza:
Thank you!

PR #307, opened 2015-06-22
fix docker stats for docker 1.7.0

slene:
@fsouza 

docker 1.7.0 stats stream api response `ContentType: text/plain` when stream=1. So I make a fix for it.

https://github.com/docker/docker/pull/13911


fsouza:
Thank you!

PR #309, opened 2015-06-23
prevent 'nil pointer dereference'

ttyh061:
Hi Fsouza:

Please take a look at this PR,

thanks,
Holden


fsouza:
Hi @ttyh061, thanks for contributing, but because the way defer works, I can't merge this the way it is. CancelRequest must be called before Body.Close, so the proper fix would require you to check whether ``err != nil`` after the ``clientconn.Do`` call and the ``HTTPClient.Do``.

This is more verbose, but will work fine.

ttyh061:
Hi fsouza:

Is this commit ok ?

fsouza:
Yes, thank you!

Issue #311, opened 2015-06-25
Dead lock in docker events not resolved in latest version


slene:
@fsouza https://github.com/fsouza/go-dockerclient/issues/203 not fix this bug

```
goroutine 7 [semacquire, 1 minutes]:
sync.(*RWMutex).Lock(0xc2080119d0)
	/usr/local/go/src/sync/rwmutex.go:87 +0xa4
github.com/fsouza/go-dockerclient.(*eventMonitoringState).updateLastSeen(0xc2080119d0, 0xc20809ed00)
	/Users/slene/src/github.com/fsouza/go-dockerclient/event.go:242 +0x33
github.com/fsouza/go-dockerclient.(*eventMonitoringState).monitorEvents(0xc2080119d0, 0xc20806a280)
	/Users/slene/src/github.com/fsouza/go-dockerclient/event.go:185 +0x245
created by github.com/fsouza/go-dockerclient.(*eventMonitoringState).enableEventMonitoring
	/Users/slene/src/github.com/fsouza/go-dockerclient/event.go:148 +0x16f

goroutine 95 [semacquire, 1 minutes]:
sync.(*RWMutex).RLock(0xc2080119d0)
	/usr/local/go/src/sync/rwmutex.go:36 +0x5f
github.com/fsouza/go-dockerclient.(*eventMonitoringState).isEnabled(0xc2080119d0, 0x0)
	/Users/slene/src/github.com/fsouza/go-dockerclient/event.go:219 +0x34
github.com/fsouza/go-dockerclient.func·006(0xc2080642d0, 0xc2080c4000)
	/Users/slene/src/github.com/fsouza/go-dockerclient/event.go:302 +0x353
created by github.com/fsouza/go-dockerclient.(*Client).eventHijack
	/Users/slene/src/github.com/fsouza/go-dockerclient/event.go:307 +0x625

goroutine 58206 [semacquire, 1 minutes]:
sync.(*RWMutex).RLock(0xc2080119d0)
	/usr/local/go/src/sync/rwmutex.go:36 +0x5f
github.com/fsouza/go-dockerclient.(*eventMonitoringState).isEnabled(0xc2080119d0, 0x0)
	/Users/slene/src/github.com/fsouza/go-dockerclient/event.go:219 +0x34
github.com/fsouza/go-dockerclient.(*eventMonitoringState).sendEvent(0xc2080119d0, 0xc20809eb80)
[...]

PR #312, opened 2015-06-25
fix events deadlock

slene:
https://github.com/fsouza/go-dockerclient/issues/311


slene:
@fsouza please review

Execute follow:

```
1 RLock   // will block Lock
2 Lock    // will block RLock and blocked by RLock
1 RLock   // blocked by Lock
1 RUnlock // wont execute, dead lock
```

fsouza:
Makes sense. Nice catch, thank you!

PR #315, opened 2015-06-28
add UTSMode to HostConfig

imikushin:
add UTSMode to HostConfig

to support docker 1.7


ibuildthecloud:
+1 

We very much need this so that we can drop some custom Docker patches in RancherOS that deal with propagating host name changes from the DHCP client.

fsouza:
Thank you!

PR #319, opened 2015-07-09
Issue http.ReadResponse hanging when running daemon with unix socket

liron-l:
On several occasions, the stream method hanged for specific container
(the scenario also repro in the docker CLI when running docker stats on
a specific container).
I tracked this error to the http.ReadResponse method but didn't find the
exact root cause in the docker daemon.
Unfortunately, we can't just 'kill' the go-routines that wait for the
stream to arrive, so this code change contains a simple cancellation
mechanism using a timer.
Please let me know if you can think of another way of solving it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"github.com/fsouza/go-dockerclient/vendor/github.com/docker/docker/pkg/stdcopy"
+	"time"
 )
 
 const userAgent = "go-dockerclient"
@@ -384,6 +385,8 @@ type streamOptions struct {
 	in             io.Reader
 	stdout         io.Writer
 	stderr         io.Writer
+	// timeout is the inital connection timeout
+	timeout time.Duration
 }
 
 func (c *Client) stream(method, path string, streamOptions streamOptions) error {
@@ -427,7 +430,17 @@ func (c *Client) stream(method, path string, streamOptions streamOptions) error 
 		if err != nil {
 			return err
 		}
+
+		// ReadResponse may hang if server does not replay
+		if streamOptions.timeout > 0 {
+			dial.SetDeadline(time.Now().Add(streamOptions.timeout))
+		}
+
 		if resp, err = http.ReadResponse(breader, req); err != nil {
+			// Cancel timeout for future I/O operations
+			if streamOptions.timeout > 0 {
+				dial.SetDeadline(time.Time{})
+			}
 			if strings.Contains(err.Error(), "connection refused") {
 				return ErrConnectionRefused
 			}
--- a/container_test.go
+++ b/container_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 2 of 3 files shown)


liron-l:
@fsouza can someone please review  ?

client.go
```diff
@@ -427,6 +430,20 @@ func (c *Client) stream(method, path string, streamOptions streamOptions) error
 		if err != nil {
 			return err
 		}
+
+		// ReadResponse may hang if server does not replay
+		connected := make(chan bool, 2)
```
peter-edge: is connected intended to be used? i know this is the general pattern, just could use some explanation

fsouza:
Hi @liron-l, sorry about the delay, I was in a conference for the last week. I will have a look in the code.

client.go
```diff
@@ -427,6 +430,20 @@ func (c *Client) stream(method, path string, streamOptions streamOptions) error
...
+		if streamOptions.timeout > 0 {
+			go func() {
+				select {
+				case <-time.After(streamOptions.timeout):
```
fsouza: Why not having just the receive and the dial.Close? Just like below:

```
<-time.After(streamOptions.timeout)
dial.Close()
```

You have a receive operation on connected, but no sends nor closes, so I can't really tell why connected exists.

fsouza:
This is a hard timeout, right? Can we have a better solution than a hard timeout for the operation? If the answer is no, could you document that this is a hard timeout for the whole operation?

Thank you very much!

liron-l:
@fsouza, @peter-edge thanks for your comments. I found a much better solution for socket timeouts using SetDeadline (it will throw a native timeout error when expired).
Thanks,
Liron

fsouza:
@liron-l such a clever solution, thank you!

This is a hard timeout, could we have an idle timeout instead?

liron-l:
Got it @fsouza. Using this pattern indeed creates hard-timeout (on future I/O ops). 
Since the connection obj (dial) is only available in the first connection scope, I think it's safe to just cancel the timeout after successful connection (so all future I/O are safe). Let me know what you think.

fsouza:
Cool, this should address the issue! Could you add a test?

Something like a request with the timeout of 100ms trying to connect to an endpoint that sleeps 1s on Accept.

liron-l:
Good idea @fsouza, I've added a test for the unix socket timeout scenario. In previous checkins, one of the UT failed (not related to my checkin), can you please check it out?

container_test.go
```diff
@@ -1585,6 +1586,39 @@ func TestTopContainerWithPsArgs(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	received := false
+	defer l.Close()
+	go func() {
+		l.Accept()
+		received = true
+		time.Sleep(time.Millisecond * 250)
+		defer l.Close()
```
fsouza: Could you remove the defer, please?

fsouza:
Thank you very much! Just a small change and it's good to merge!

liron-l:
Thanks @fsouza, removed.

fsouza:
Thank you!

PR #321, opened 2015-07-10
Updating volumes-from to be []string, not string

bfosberry:


peter-edge:
This would break a lot of existing dependent code I would think, no?

bfosberry:
Most likely! A better solution would be to use []string internally and do
remote api version detection, or abstract the type. Either way its going to
break some stuff. Hopefully we can get a plan in place to minimise damage.

peter-edge:
In general, I don't think this repo has ever done a breaking change like that, is that correct, @fsouza? If we wanted to go that route, there should be more complete handling in place/implemented

bfosberry:
The issue is its currently broken for me

peter-edge:
This PR does more closely match the docker remote API https://docs.docker.com/reference/api/docker_remote_api_v1.19/#create-a-container, I think I have had this working in the past with CSV values but I could be wrong, that might have been another field

bfosberry:
This whole thing smells of docker making a sublte breaking change

peter-edge:
Ah ya, I did have it working haha https://github.com/peter-edge/go-dockerutils/blob/master/dockerutils.go#L174

fsouza:
Hi @bfosberry @peter-edge, maybe Docker moved VolumesFrom from Config to HostConfig, so the value in Config doesn't do anything, because it doesn't exist in Docker's config: https://github.com/docker/docker/blob/master/runconfig/config.go#L117-L140.

They did it with security options, memory, swap and CPU shares as well. There are probably other cases.

I'd rather just document that VolumesFrom in go-dockerclient is deprecated, and users should specify it in the HostConfig type.

peter-edge:
SGTM, ping @bfosberry do you want to do that in this PR?

peter-edge:
Can we close this for now?

fsouza:
@peter-edge yes, I'm closing it. Thanks!

bfosberry:
I'll try that out on our side, thanks!

PR #323, opened 2015-07-11
Fix nil panic when fails to get response from unix socket

styner32:
From: https://github.com/fsouza/go-dockerclient/blob/master/client.go#L347
```
if protocol == "unix" {
		dial, err := net.Dial(protocol, address)
		if err != nil {
			return nil, -1, err
		}
		defer dial.Close()
		breader := bufio.NewReader(dial)
		err = req.Write(dial)
		if err != nil {
			return nil, -1, err
		}
		resp, err = http.ReadResponse(breader, req)
	} else {
		resp, err = c.HTTPClient.Do(req)
	}
	if err != nil {
		if strings.Contains(err.Error(), "connection refused") {
			return nil, -1, ErrConnectionRefused
		}
		return nil, -1, err
	}
	defer resp.Body.Close()
```

When `http.ReadResponse(breader, req)` return error, since `err` is re-declared on `dial, err := net.Dial(protocol, address)`, `err` is still nil on `if err != nil {`. So it panics on `defer resp.Body.Close()` because `resp` is nil. 

I fixed not to re-declare in `if` block, so it assigns the error to `err` properly.

### DIFF
--- a/client.go
+++ b/client.go
@@ -345,7 +345,8 @@ func (c *Client) do(method, path string, doOptions doOptions) ([]byte, int, erro
 	protocol := c.endpointURL.Scheme
 	address := c.endpointURL.Path
 	if protocol == "unix" {
-		dial, err := net.Dial(protocol, address)
+		var dial net.Conn
+		dial, err = net.Dial(protocol, address)
 		if err != nil {
 			return nil, -1, err
 		}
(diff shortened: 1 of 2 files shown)


peter-edge:
cute :)

peter-edge:
LGTM @fsouza

fsouza:
Nice catch, thanks!

PR #324, opened 2015-07-13
Fix deadlock in disableEventMonitoring():eventState.Wait()

orivej:
Currently the following pseudo-example deadlocks when docker daemon is not running:

     events := make(chan *docker.APIEvents)
     dc, err := docker.NewClient() // no error here
     err = dc.AddEventListener(events) // no error here
     for ev := range events {
              // unreachable
     }
     // unreachable

The client hangs in monitorEvents→connectWithRetry/terminate→disableEventMonitoring→eventState.Wait


fsouza:
Nice catch, thank you!

PR #329, opened 2015-07-13
Update client.go

### LINKED ISSUE #328: Repeated TLS connections to docker leak file handles
I have an application that schedules periodic calls to several docker servers. It wakes up, connects to the server via TLS, issues a few commands, and then goes back to sleep. I noticed after running for a while, it would start to use a lot of file handles. I have distilled the bug down to the following code example. It will leak one handle per second. I tried the same thing without TLS, and the bug does not happen.

```
package main

import (
    "time"
    "fmt"
    "github.com/fsouza/go-dockerclient"
    "io/ioutil"
)

func main() {
    cert, _ := ioutil.ReadFile("cert.pem")
    key, _ := ioutil.ReadFile("key.pem")

    for {
        client, err := docker.NewTLSClientFromBytes("http://docker-host:2376", cert, key, nil)
        fmt.Printf("connect: %v\n", err)

        err = client.Ping()
        fmt.Printf("ping: %v\n", err)

        time.Sleep(1 * time.Second)
    }
}
```

tdterry:
Fixes #328 

I looked at the TLS client code and found the problem. At the bottom of NewVersionedTLSClientFromBytes() you create a Transport with the TLS configuration. Transport uses keep-alive by default and will leave that connection open waiting for another request, but since each call to NewTLSClient() creates another new Transport, the connections are never reused.

The simplest fix is to disable keep-alive in the TLS transport. This has no performance impact over the current implementation because it wasn't using keep-alive anyway. A more performant solution would be to cache the TLS transport based on `endpoint` to be reused on subsequent calls.

### DIFF
--- a/client.go
+++ b/client.go
@@ -245,7 +245,7 @@ func NewVersionedTLSClientFromBytes(endpoint string, certPEMBlock, keyPEMBlock, 
 		tlsConfig.RootCAs = caPool
 	}
 	tr := &http.Transport{
-		TLSClientConfig: tlsConfig,
+		TLSClientConfig:   tlsConfig,
 		DisableKeepAlives: true,
 	}
 	if err != nil {
--- a/client.go
+++ b/client.go
@@ -246,6 +246,7 @@ func NewVersionedTLSClientFromBytes(endpoint string, certPEMBlock, keyPEMBlock, 
 	}
 	tr := &http.Transport{
 		TLSClientConfig: tlsConfig,
+		DisableKeepAlives: true,
 	}
 	if err != nil {
 		return nil, err


fsouza:
@tdterry I don't think this is the right fix. Disabling keep alive doesn't seems right to me. We need further investigation on the issue. Does the behavior reported in #328 happen only when using a TLS-enabled client?

tdterry:
Yes. It only happens with TLS, and only when you create multiple NewTLSClient() to the same machine. In order to have keep-alive, you need to keep the transport around, which means it needs to be stored outside the docker.Client struct. I can update the PR to do that.

fsouza:
@tdterry we should keep the transport inside the Client instance. I wonder why you need to create multiple client instances. Could you share more of your use case? Thanks!

peter-edge:
@tdterry any thoughts on this? :)

client.go
```diff
@@ -245,7 +245,8 @@ func NewVersionedTLSClientFromBytes(endpoint string, certPEMBlock, keyPEMBlock,
 		tlsConfig.RootCAs = caPool
 	}
 	tr := &http.Transport{
-		TLSClientConfig: tlsConfig,
+		TLSClientConfig:   tlsConfig,
+		DisableKeepAlives: true,
```
liggitt: careful, this will make reuse of a single transport much more expensive (re-establishing a TLS connection for every request)

peter-edge:
@fsouza should we close this? we haven't had any input for a while

tdterry:
I'm fine either way. I ended up changing my code to only have a single instance of the Client per host.

fsouza:
@peter-edge yeah, thanks!

PR #330, opened 2015-07-16
correct behavior of events monitor

slene:
Hi, I put some fix of docker client.

1. when connect docker failed should terminate events monitor (now is blocked and deadlock)
2. when docker disconnect should terminate events monitorr (now is blockedd and deadlock)

@fsouza


fsouza:
Thanks!

PR #336, opened 2015-07-27
Add MacAddress to NetworkSettings

paulbellamy:
It's both in Config, and in NetworkSettings.

### DIFF
--- a/container.go
+++ b/container.go
@@ -130,6 +130,7 @@ type PortMapping map[string]string
 type NetworkSettings struct {
 	IPAddress   string                 `json:"IPAddress,omitempty" yaml:"IPAddress,omitempty"`
 	IPPrefixLen int                    `json:"IPPrefixLen,omitempty" yaml:"IPPrefixLen,omitempty"`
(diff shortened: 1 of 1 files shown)


peter-edge:
Can you just link quickly to the location this is in github.com/docker/docker? There's been a duplication problem lately with these adds

fsouza:
From what I know Docker, I suspect the field in the Config struct is for customizing the Mac address on container creation, and the field in NetworkSettings is for checking it later (in InspectContainer, for example).

@paulbellamy can you confirm that this is the case?

Thanks for contributing!

paulbellamy:
I'm going based on the api documentation here: https://docs.docker.com/reference/api/docker_remote_api_v1.19/#inspect-a-container

fsouza:
Indeed it exists in both places, as I've suspected before:

- Config: https://github.com/docker/docker/blob/967bc13ae68939c616033710f7695993aa4f5208/runconfig/config.go#L137
- NetworkSettings: https://github.com/docker/docker/blob/967bc13ae68939c616033710f7695993aa4f5208/daemon/network/settings.go#L24

I will merge this pull request, thank you very much for contributing!

peter-edge:
ugh docker :)

PR #337, opened 2015-07-29
Add 'LogPath' property in the 'container' struct

fpgeek:
The 'LogPath' property was added since v1.18. Please add 'LogPath' property in the 'container' struct


fsouza:
Thank you!

PR #341, opened 2015-08-03
godep save -r ./...

freeformz:


fsouza:
@freeformz did you open this pull request just to link the diff in tools/godep#224?

peter-edge:
@fsouza I'm not completely sure on the background but I don't think this should be merged - we use a different vendoring tool, and the reason we didn't use godep was because it is hard to use go-dockerclient as a library across go versions if godep is used. Regardless, this just doubles the vendoring in here with what we do with party.

freeformz:
@fsouza sorry about the noise.

fsouza:
@freeformz no problem, I just wanted to get more context!

freeformz:
@fsouza I would like to get some more clarity around your statements about godep vs party though at some point. But a private email for that is probably better (edward (at) heroku (dot) com) if you have a few minutes.

PR #342, opened 2015-08-04
removed non-existant symlink file from testing/data.

elevran:
The symlink file points to a non-existing target and is not used. May be leftover from local developer directory?


fsouza:
@elevran it's used in for testing purpose, as you can see in the test output: https://travis-ci.org/fsouza/go-dockerclient/jobs/74094588.

PR #343, opened 2015-08-04
Rudimentary docker network support

dcbw:
Start adding support for docker's network API (which is just a passthrough to libnetwork's API).

### DIFF
--- a/container.go
+++ b/container.go
@@ -128,13 +128,23 @@ type PortMapping map[string]string
 
 // NetworkSettings contains network-related information about a container
 type NetworkSettings struct {
-	IPAddress   string                 `json:"IPAddress,omitempty" yaml:"IPAddress,omitempty"`
-	IPPrefixLen int                    `json:"IPPrefixLen,omitempty" yaml:"IPPrefixLen,omitempty"`
-	MacAddress  string                 `json:"MacAddress,omitempty" yaml:"MacAddress,omitempty"`
-	Gateway     string                 `json:"Gateway,omitempty" yaml:"Gateway,omitempty"`
-	Bridge      string                 `json:"Bridge,omitempty" yaml:"Bridge,omitempty"`
-	PortMapping map[string]PortMapping `json:"PortMapping,omitempty" yaml:"PortMapping,omitempty"`
-	Ports       map[Port][]PortBinding `json:"Ports,omitempty" yaml:"Ports,omitempty"`
+	IPAddress              string                 `json:"IPAddress,omitempty" yaml:"IPAddress,omitempty"`
+	IPPrefixLen            int                    `json:"IPPrefixLen,omitempty" yaml:"IPPrefixLen,omitempty"`
+	MacAddress             string                 `json:"MacAddress,omitempty" yaml:"MacAddress,omitempty"`
+	Gateway                string                 `json:"Gateway,omitempty" yaml:"Gateway,omitempty"`
+	Bridge                 string                 `json:"Bridge,omitempty" yaml:"Bridge,omitempty"`
+	PortMapping            map[string]PortMapping `json:"PortMapping,omitempty" yaml:"PortMapping,omitempty"`
+	Ports                  map[Port][]PortBinding `json:"Ports,omitempty" yaml:"Ports,omitempty"`
+	NetworkID              string                 `json:"NetworkID,omitempty" yaml:"NetworkID,omitempty"`
+	EndpointID             string                 `json:"EndpointID,omitempty" yaml:"EndpointID,omitempty"`
+	SandboxKey             string                 `json:"SandboxKey,omitempty" yaml:"SandboxKey,omitempty"`
+	GlobalIPv6Address      string                 `json:"GlobalIPv6Address,omitempty" yaml:"GlobalIPv6Address,omitempty"`
+	GlobalIPv6PrefixLen    int                    `json:"GlobalIPv6PrefixLen,omitempty" yaml:"GlobalIPv6PrefixLen,omitempty"`
+	IPv6Gateway            string                 `json:"IPv6Gateway,omitempty" yaml:"IPv6Gateway,omitempty"`
+	LinkLocalIPv6Address   string                 `json:"LinkLocalIPv6Address,omitempty" yaml:"LinkLocalIPv6Address,omitempty"`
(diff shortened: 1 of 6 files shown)


peter-edge:
Hey, just for reference, can you link the github.com/docker/docker corresponding code/docs? If you don't have time, I can check it out soon.

network.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// Service represents a service.
+type Service struct {
+	ID string `json:"id"`
+}
+
+// ListNetworks returns all networks.
+func (c *Client) ListNetworks() ([]Network, error) {
+	path := "/networks"
```
peter-edge: nit: might as well just inline this :)

network.go
```diff
@@ -0,0 +1,76 @@
...
+	err = json.Unmarshal(body, &networks)
+	if err != nil {
+		return nil, err
+	}
+	return networks, nil
+}
+
+// NetworkInfo returns information about a network by its ID.
+func (c *Client) NetworkInfo(id string) (*Network, error) {
+	path := "/networks/" + id
```
peter-edge: same comment

network.go
```diff
@@ -0,0 +1,76 @@
...
+
+// ListNetworks returns all networks.
+func (c *Client) ListNetworks() ([]Network, error) {
+	path := "/networks"
+	body, _, err := c.do("GET", path, doOptions{})
+	if err != nil {
+		return nil, err
+	}
+	var networks []Network
+	err = json.Unmarshal(body, &networks)
```
peter-edge: scope this (and below):

```
if err := json.Unmarshal(body, &networks); err != nil {
  return nil, err
}
```

network.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &network, nil
+}
+
+// NoSuchNetwork is the error returned when a given network does not exist.
+type NoSuchNetwork struct {
+	ID  string
+	Err error
```
peter-edge: do you ever use Err? it doesn't look like it here.

network.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (err *NoSuchNetwork) Error() string {
+	if err.Err != nil {
+		return err.Err.Error()
+	}
+	return "No such network: " + err.ID
```
peter-edge: nit: i usually like `fmt.Sprintf("foo: %s", err.ID)` but no biggie

peter-edge:
Also overall: needs some testing :)

container.go
```diff
@@ -135,6 +135,8 @@ type NetworkSettings struct {
 	Bridge      string                 `json:"Bridge,omitempty" yaml:"Bridge,omitempty"`
 	PortMapping map[string]PortMapping `json:"PortMapping,omitempty" yaml:"PortMapping,omitempty"`
 	Ports       map[Port][]PortBinding `json:"Ports,omitempty" yaml:"Ports,omitempty"`
+	NetworkID   string                 `json:"NetworkID,omitempty" yaml:"NetworkID,omitempty"`
```
peter-edge: just making sure - is it `json:"NetworkID"` or `json:"NetworkId"` for docker?

dcbw: It's "NetworkID":

        "MacAddress": "02:42:ac:11:2a:02",
        "NetworkID": "b6408b405133c5475bfde2ad383b80732326e1fb9bc713f6586552836ba2fdd5",
        "PortMapping": null,
        "Ports": {},
        "SandboxKey": "/var/run/docker/netns/23659bc4a779",
        "SecondaryIPAddresses": null,
        "SecondaryIPv6Addresses": null

network.go
dcbw: I mainly copied NoSuchContainer, which also has Err and doesn't appear to use it.  I'm happy to take it out of NoSuchNetwork, but should I be consistent with NoSuchContainer, or should I remove the unused bit?

peter-edge: Remove both, thanks :)

container.go
peter-edge: SGTM

network.go
dcbw: Sure.  I removed Service too, since it's currently unused.

dcbw:
Repushed with some fixups.

peter-edge:
Thanks! Still needs to address the string inlining, linking to docker/docker code, and tests :)

dcbw:
FWIW the JSON looks like this:

dcbw:
FWIW the JSON for Network looks like this:

&{Name:purple ID:2585945e1bb6f36514dde1c563dd8d078bc058d9a4953a0fd8780b3b5237e983 Type:cni Endpoints:[]}

dcbw:
Repushed with inlined string and added links to docker code.  No documentation yet on the API that I can find (not surprising, it's really new).  Still have to do some tests; but does the code look OK so far now?

fsouza:
@peter-edge thanks for reviewing this!

@dcbw except for the lack of tests, I think the code looks good for now. Regarding the Docker documentation issue, are these APIs available in Docker 1.8? If so, they're probably going to add some docs soon.

dcbw:
@fsouza these APIs are available in docker 1.8 if you enable the EXPERIMENTAL flag when building docker.  It seems very likely they'll get turned on by default in the near future though.

peter-edge:
@fsouza i think it's good to merge but could use at least some base for testing so that when this code is inevitably expanded, other OSS contributors know where to add tests. plus some documentation about the need to do EXPERIMENTAL/1.8 etc. @dcbw are you willing to do this? :)

dcbw:
@peter-edge yes, I'm willing to work on the tests in the next few days.

fsouza:
Cool, we can wait for the tests then!

@dcbw thank you!

dcbw:
Ok, added some testcases, do they look complete enough?  Also added a note to the README about the network API and DOCKER_EXPERIMENTAL, and added some more fields to the NetworkSettings struct that are present in docker 1.8.

peter-edge:
Looks great! @fsouza

fsouza:
I'm missing the tests for the CreateNetwork method. @dcbw any special reason for not having tests for this method?

Thank you! And sorry for the late response.

dcbw:
@fsouza no, just an oversight, sorry!  Added the client-side CreateNetwork tests and repushed.

fsouza:
@dcbw thank you!

PR #344, opened 2015-08-05
Fixing auth bug

bfosberry:


bfosberry:
Dang, need to strip out other commit, hold up

PR #346, opened 2015-08-06
listening should be in a goroutine

tonicbupt:
When I use function `Stats` I meet one problem. If the container dies(just exited, not removed yet) immediately after it starts, `Stats` will be blocked, now if I want it to stop, I send signal `true` to channel `opts.Done`, however the code never runs into `select`. So I think as a observer this should be running in a separated goroutine.

### DIFF
--- a/container.go
+++ b/container.go
@@ -726,11 +726,15 @@ func (c *Client) Stats(opts StatsOptions) (retErr error) {
 		close(errC)
 	}()
 
+	quit := make(chan struct{})
+	defer close(quit)
 	go func() {
 		// block here waiting for the signal to stop function
 		select {
 		case <-opts.Done:
 			readCloser.Close()
+		case <-quit:
+			return
 		}
 	}()
 
--- a/container.go
+++ b/container.go
@@ -726,6 +726,14 @@ func (c *Client) Stats(opts StatsOptions) (retErr error) {
 		close(errC)
 	}()
 
+	go func() {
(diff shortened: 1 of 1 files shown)


container.go
```diff
@@ -726,6 +726,14 @@ func (c *Client) Stats(opts StatsOptions) (retErr error) {
...
 	}()
 
+	go func() {
+		// block here waiting for the signal to stop function
+		select {
+		case <-opts.Done:
+			readCloser.Close()
+		}
+	}()
+
```
fsouza: I'd like to be able to prevent this goroutine from leaking. I.e.: if done never gets closed or a value, it will block forever. You could create another channel, and add another case to the select statement, something like:

```go
quit := make(chan struct{})
defer close(quit)
go func() {
	select {
	case <-opts.Done:
		readCloser.Close()
	case <-quit:
		return
	}
}
```

Does it make sense?

container.go
```diff
@@ -726,6 +726,18 @@ func (c *Client) Stats(opts StatsOptions) (retErr error) {
...
+	go func() {
+		// block here waiting for the signal to stop function
+		select {
+		case <-opts.Done:
+			readCloser.Close()
+		case <-quit:
+			return
+		}
+	}()
+
```
tonicbupt: @fsouza done... I just knew closing a channel can still go to `<-quit` ...

fsouza:
Thank you!

PR #347, opened 2015-08-07
determinism in hijack error returns

codekitchen:
Previously, if both input and output streams were specified, it was random which
error was returned. So if the output write consistently fails, you'd still
sometimes get a nil error back, because sometimes the input error chan value happened to
get selected first (and was nil).

Now, we explicitly wait for both errors, and return any non-nil error,
preferring the input error.

Possibly it'd make sense to just return both errors? But that'd be an API change.

### DIFF
--- a/client.go
+++ b/client.go
@@ -28,10 +28,11 @@ import (
 	"strconv"
 	"strings"
 
+	"time"
+
 	"github.com/fsouza/go-dockerclient/external/github.com/docker/docker/opts"
 	"github.com/fsouza/go-dockerclient/external/github.com/docker/docker/pkg/homedir"
 	"github.com/fsouza/go-dockerclient/external/github.com/docker/docker/pkg/stdcopy"
-	"time"
 )
 
 const userAgent = "go-dockerclient"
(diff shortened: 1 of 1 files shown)


codekitchen:
Oh hey Peter, it's been a while.

Our use case is a Writer that returns an error if too much data is written. Without this bug fix, our test case gets no error back from `AttachToContainer` about half the time, even though the Writer actually returned an error.

I can't see any downside other than what I mentioned about it only returning one error if both sides errored, but that was true before as well.

peter-edge:
Ya man long time :) Google's doing their own protobuf for ruby now, you see that? Heh.

Ya just gonna let @fsouza chime in.

fsouza:
@codekitchen thanks for contributing. As far as I remember, we had some issues with errIn returning before the output goroutine exits, but since you first get the errors and then check them, I don't think we will run in the same troubles now.

I will run some other tests here to ensure that everything is fine with this change and merge it in case everything goes fine.

@peter-edge thanks for looking into it! :)

fsouza:
Everything worked fine, thanks!

Merging it.

codekitchen:
Sweet, thanks for the quick review.

Issue #348, opened 2015-08-11
Import path rewriting and GO15VENDOREXPERIMENT do not mix


pwaller:
Please see https://github.com/golang/go/issues/12111.

In my case, `vendor.org/p` is `fsouza/go-dockerclient`. The basic problem is that I cannot import fsouza/go-dockerclient with `GO15VENDOREXPERIMENT` set (and I want to set it in all of my projects!). I think the only solution may be to avoid the rewriting or rename the `/vendor/` directory in this project.

pwaller:
A recommendation from the linked issue is to use a directory called `internal` instead of `vendor`. (Though it is preferred to use the new vendoring way wherever it is possible. We might have to wait for GO15VENDOREXPERIMENT to become standard first though).

peter-edge:
We already re-named it to external, see https://github.com/fsouza/go-dockerclient/pull/339 and https://github.com/fsouza/go-dockerclient/issues/338

pwaller:
@peter-edge aha! Apologies for the noise, I missed the fact I was looking at an out of date module.

peter-edge:
All good man haha

PR #349, opened 2015-08-11
Add a note about vendoring for go1.5 in the readme

peter-edge:
We're about to get a barrage of comments about this. Kubernetes is now updated but there's going to be a lot of users that are not yet and they're going to come to the project and wonder what is doing on. Adding this documentation for Now. Also re: https://github.com/fsouza/go-dockerclient/issues/348


fsouza:
Thank you!

PR #351, opened 2015-08-12
defer close if streamOptions.std* is an io.Closer

wplu:
When passing in an io.Closer for OutputStream, like a io.PipeWriter, it would be nice to receive indication when that stream is closed.

### DIFF
--- a/client.go
+++ b/client.go
@@ -462,9 +462,13 @@ func (c *Client) stream(method, path string, streamOptions streamOptions) error 
 	address := c.endpointURL.Path
 	if streamOptions.stdout == nil {
 		streamOptions.stdout = ioutil.Discard
+	} else if t, ok := streamOptions.stdout.(io.Closer); ok {
+		defer t.Close()
 	}
 	if streamOptions.stderr == nil {
 		streamOptions.stderr = ioutil.Discard
+	} else if t, ok := streamOptions.stderr.(io.Closer); ok {
+		defer t.Close()
 	}
 	if protocol == "unix" {
 		dial, err := net.Dial(protocol, address)


fsouza:
Hi @wplu, thanks for contributing! I'd rather not take this approach, as I believe that whoever creates the ReadCloser should be responsible for closing it. As we ask for a Reader only, we should be only reading from it, and let the owner of the ReadCloser close it.

Does it make sense?

Thank you.

wplu:
Hi @fsouza, that would be ok, which i've been doing as a workaround.  However, we're not talking about Reads,  we are talking about setting OutputStream to an io.PipeWriter.  If we follow the example here:
  http://nathanleclaire.com/blog/2014/07/19/demystifying-golangs-io-dot-reader-and-io-dot-writer-interfaces/
but change the s.api.AttachToContainer() call to s.api.PullImage(), then the code never leaves the loop for scanner.Scan() unless we call p.Close() after the s.api.PullImage() finishes.  It would be more natural to get an io.EOF when the PullImage() is finished instead of having to mak an additional call to p.Close().

fsouza:
I see, thanks for the clarification, and sorry for the confusion about the object kind, but I still think that it's the caller responsibility to close the stream. So, in the post you linked, you'd have something like:

```go
go func() {
    defer w.Close()
    s.api.AttachToContainer(options)
}()
```

wplu:
Yes, sorry I wasn't clear.  I would only have the defer w.Close() for PullImage(), but not for AttachContainer()

You have a point about the caller responsibility to close the stream.  However, closing the write side of the stream is generally the writer's responsibility to indicate EOF.

fsouza:
@wplu that's a good point, thanks for bringing it. Could you document in all methods that call stream that when the Output and Error streams are ReadClosers, the close method will be called to signal EOF?

Thanks!

wplu:
@fsouza - will do!

PR #355, opened 2015-08-21
Use the file as a io.Reader

gregdel:
No need to read the whole file in memory and use a buffer, a file is already a
io.Reader


fsouza:
Nice catch, thanks!

PR #356, opened 2015-08-21
return error when attempting to start container with empty id

daniel-garcia:
When attempting to start a container with "" as the container id, no error is returned. It's not clear that upstream can catch such an error since it is not a valid URL that gets generated.

### DIFF
--- a/container.go
+++ b/container.go
@@ -482,6 +482,9 @@ type HostConfig struct {
 //
 // See http://goo.gl/iM5GYs for more details.
 func (c *Client) StartContainer(id string, hostConfig *HostConfig) error {
+	if id == "" {
+		return &NoSuchContainer{ID: id, Err: fmt.Errorf("empty container id")}
+	}
 	path := "/containers/" + id + "/start"
 	_, status, err := c.do("POST", path, doOptions{data: hostConfig, forceJSON: true})
 	if status == http.StatusNotFound {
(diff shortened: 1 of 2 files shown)


daniel-garcia:
After deeper inspection of the code, there are many calls that suffer from the same problem. Should all of them be addressed in this PR and this the right approach, @fsouza ?

container.go
```diff
@@ -482,6 +482,9 @@ type HostConfig struct {
 //
 // See http://goo.gl/iM5GYs for more details.
 func (c *Client) StartContainer(id string, hostConfig *HostConfig) error {
+	if id == "" {
+		return &NoSuchContainer{ID: id, Err: fmt.Errorf("empty container id")}
```
peter-edge: I think this should be a new error type instead of using NoSuchContainer - NoSuchContainer is used to denote when you actually have an invalid container id, just just had an empty parameter string.

peter-edge:
I'm not sure if we should go down this route or not - I think we either should verify parameters on all functions, or do not verify at all (which is basically the current case).

fsouza:
Hi @daniel-garcia, thanks for contributing! Doesn't the Docker API return no error at all? Or are we not handling it properly?

I prefer to delegate the error handling to the Docker API. If users want to save an API call in case the container ID is empty, they should wrap the client call with an if. But I'm surprised that the current code doesn't return an error on StartContainer when the id is empty.

fsouza:
Just tested locally and it fails because Docker returns a 301, and then a 404, but go-dockerclient doesn't follow redirects.

fsouza:
@daniel-garcia I will close this PR, please feel free to keep the discussion going on, but I don't think we should be validating the contents of the id.

Thanks for contributing!

PR #358, opened 2015-08-26
Add Labels in APIContainers

Soulou:
Available since API 1.18

https://docs.docker.com/reference/api/docker_remote_api_v1.18/#list-containers


fsouza:
Thank you!

PR #367, opened 2015-09-03
Add VolumeDriver parameter, new in Docker 1.8

bboreham:
This is analogous to the `--volume-driver` parameter in the Docker CLI


fsouza:
Thank you!

Issue #368, opened 2015-09-04
Avoid pollution of user's filesystems with dangling symlink


flowlo:
This repository contains [a symlink to a non-existent target](https://github.com/fsouza/go-dockerclient/blob/master/testing/data/symlink). I don't want dangling symlinks on my root filesystem.

Also, it breaks deploying a Google App Engine Managed VM that imports this with [`aedeploy`](https://github.com/golang/appengine/blob/master/cmd/aedeploy/aedeploy.go) because copying the symlink (obviously) fails. [This is the recommended way to deploy those apps.](https://groups.google.com/forum/#!topic/google-appengine-go/D7eytg7btr8)

I would appreaciate if you could get rid of that symlink. As far as I see it is only there for testing, so I assume it could be created and deleted on demand.

flowlo:
Sorry about referencing this from two commits, I forgot to rebase first :flushed:

PR #369, opened 2015-09-04
testing: non-existent symlink on demand

flowlo:
Fixes fsouza/go-dockerclient#368

### DIFF
--- a/build_test.go
+++ b/build_test.go
@@ -35,6 +35,16 @@ func TestBuildImageMultipleContextsError(t *testing.T) {
 func TestBuildImageContextDirDockerignoreParsing(t *testing.T) {
 	fakeRT := &FakeRoundTripper{message: "", status: http.StatusOK}
 	client := newTestClient(fakeRT)
+
+	if err := os.Symlink("doesnotexist", "testing/data/symlink"); err != nil {
+		t.Errorf("error creating symlink on demand: %s", err)
+	}
+	defer func() {
+		if err := os.Remove("testing/data/symlink"); err != nil {
+			t.Errorf("error removing symlink on demand: %s", err)
+		}
+	}()
+
 	var buf bytes.Buffer
 	opts := BuildImageOptions{
 		Name:                "testImage",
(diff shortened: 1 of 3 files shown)


build_test.go
```diff
@@ -35,6 +35,11 @@ func TestBuildImageMultipleContextsError(t *testing.T) {
 func TestBuildImageContextDirDockerignoreParsing(t *testing.T) {
 	fakeRT := &FakeRoundTripper{message: "", status: http.StatusOK}
 	client := newTestClient(fakeRT)
+
+	if err := os.Symlink("doesnotexist", "testing/data/symlink"); err != nil {
+		t.Errorf("error creating symlink on demand: %s", err)
+	}
+
```
fsouza: There are some calls to t.Fatal in this test, so I'd prefer to have a defer just after this if:

```
defer func() {
	if err := os.Remove("testing/data/symlink"); err != nil {
		t.Errorf("error removing symlink on demand: %s", err)
	}
}()
```

Thank you!

flowlo: Oh sure!

flowlo:
There is one failing job, but I can't figure out how this correlates with my changes.

flowlo:
Ping @fsouza @peter-edge

peter-edge:
The defer will not go off if any t.Fatal function is hit. Therefore, this might not get cleaned up. That's OK, but I think you should handle it, both in the .gitignore, and check for existence before creating it.

flowlo:
Adding `testing/data/symlink` to `.gitignore` makes sense, but I am not so sure about adding logic to handle any existing file. Probably we'd have to `os.Lstat` the symlink. In this case (a) `os.Lstat` itself can fail, (b) the file could exists but not be a symlink, and (c) should we remove it?

Do you really think that this will be the case very often? I'd probably let the test error if there's a file named `testin/data/symlink` and leave the rest up to the user executing the tests.

fsouza:
@peter-edge yeah, you're right about the fatal vs defer thing :-(

@flowlo I think it's fine to merge as is, maybe we can drop Fatal usage in the future. Thanks for contributing!

PR #370, opened 2015-09-05
use net.Dialer instance to establish connections

cezarsa:
A pointer to a net.Dialer is stored in the Client struct. This allow clients to override its value the same way it's currently possible with the http.Client.


cezarsa:
Merging after talking with @fsouza IRL.

Issue #352, opened 2015-08-16
support api v1.20 new ability - upload files to container


elasti-erezh:
Note that the new api deprecates the copy files or folders from a container.

Please support the new added ability to upload/download tarred files to/from a container.

From [docker remote api v1.20](https://docs.docker.com/reference/api/docker_remote_api_v1.20/#extract-an-archive-of-files-or-folders-to-a-directory-in-a-container) :
`PUT /containers/(id)/archive` - Upload an archive of content to be extracted to an existing directory inside a container’s filesystem.

elasti-erezh:
started working on it and will send pull request when done

fsouza:
@elasti-erezh thank you!

jandre:
^^ does that work?  I need to figure out how to add a test for this.

elasti-erezh:
@jandre i've tested it vs a live container and it works well.

Don't have a test for it yet..

elasti-erezh:
@fsouza this can be closed right ? 
saw @jandre already took care of it 
Sorry for the lag and thank you

fsouza:
@elasti-erezh no problem! Thanks for opening the issue.

PR #371, opened 2015-09-06
support GET /containers/(id)/archive for Docker API >= 1.20 [#352]

elasti-erezh:


container.go
```diff
@@ -860,6 +860,31 @@ func (c *Client) PutContainerArchive(id string, opts PutContainerArchiveOptions)
 	})
 }
 
+// GetContainerArchiveOptions is the set of options that can be used when
+// downloading resources from a container.
+//
+// See https://goo.gl/KnZJDX for more details.
+type GetContainerArchiveOptions struct {
+	OutputStream io.Writer `json:"-" qs:"-"`
+	Path         string    `qs:"path"`
```
peter-edge: Is path optional or required? If required, it might be better as a parameter along with id

elasti-erezh: Yes it is mandatory. I agree its safer, the reason i left it inside the opts is to be consistent with the complement PutContainerArchive func. so i'll change them both to match.

elasti-erezh: Yes it is mandatory. I agree its safer, but the're are couple of reasons to leave it inside the opts (and maybe move id back to opts as well):
* The reason i left it inside the opts is to be consistent with code since all other funcs since get opts as params.
* Getting 'path' as param means that we won't use the queryString func formatter and 'path'  must be embedded as part of the url, this fact is also breaking consistency and beauty of code 
* Being repetitive but all funcs has some required params which being transferred as part of the opts struct and it does make since in a way that the definition and the call for those funcs are very clear and readable. Example:  call of func("123", "/file", false) its less clear then calling func(Opts{ID: "123", Path: "/file", Force: false})
Hope i convinced you :).

Can you also give your opinion on my other 'REVIEW' notes please?
Thank you in advance

peter-edge: There are other functions that have multiple parameters. Opts are used for actually optional things, so this is confusing. id definitely should not be in opts, but path I also think should not be, I would be confused by that as a user and I use go-dockerclient all the time.

peter-edge: But ya I can see the consistency argument. Eh.

container.go
```diff
@@ -860,20 +860,38 @@ func (c *Client) PutContainerArchive(id string, opts PutContainerArchiveOptions)
...
+func (c *Client) GetContainerArchive(id string, opts GetContainerArchiveOptions) error {
+	url := fmt.Sprintf("/containers/%s/archive?", id) + queryString(opts)
+
+	return c.stream("GET", url, streamOptions{
+		setRawTerminal: true,
+		stdout:         opts.OutputStream,
+	})
+}
+
+// CopyFromContainer has been DEPRECATED, please use GetContainerArchive.
```
fsouza: Please use CopyFromContainerOptions, just to make sure golint is happy! Something like:

```
// CopyFromContainerOptions has been DEPRECATED, please use GetContainerArchiveOptions along with GetContainerArchive.
```

container.go
```diff
@@ -860,20 +860,38 @@ func (c *Client) PutContainerArchive(id string, opts PutContainerArchiveOptions)
...
+	OutputStream io.Writer `json:"-" qs:"-"`
+	Path         string    `qs:"path"`
+}
+
+// REVIEW: 'GetContainerArchive' Needs a better name since there's also 'ExportContainer', confusing?!
+// 		 IMO change both PUT/GET funcs to:  'DownloadFromContainer' & 'UploadToContainer'
+
+// GetContainerArchive downloads a tar archive of files or folders in a container.
+//
+// See https://goo.gl/KnZJDX for more details.
```
fsouza: I like the idea of using DownloadFromContainer and UploadToContainer, feel free to change both methods (I know it will break the later, but I think it's fresh enough).

fsouza:
Hi @elasti-erezh, sorry about the late response and the conflict I caused, can you address the comments and fix the conflict?

Thank you!

container.go
elasti-erezh: Sure :)
Done at the following pull request

elasti-erezh:
Include updated code after review ..

elasti-erezh:
@fsouza , I've pushed force my repo with changes and it automatically closed the pull request (don't know why??) so had to reopen the issue.
Last 2 commits hold all the changes.
Thank you

fsouza:
@elasti-erezh thank you!

elasti-erezh:
Thank you.  awesome api package ;-)

PR #375, opened 2015-09-09
Don't read HTTP entire response body unnecessarily

chuyeow:
We ran into this problem where copying a large file from the container consumed a lot of memory.

I realize that the problem is because [`Client.do()` reads the entire HTTP response from the remote API](https://github.com/fsouza/go-dockerclient/blob/49c1b8c3ae297cdbddb0aad022b82c0480f92877/client.go#L422).

This is an unnecessary memory / allocation burden, so I've changed it to return the `*http.Response` and rely on callers to read / copy from / close the body as necessary.

Since I was trying to fix `CopyFromContainer`, I noticed that its test wasn't actually testing the contents of its output stream - I've fixed that too.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -127,12 +127,10 @@ func (c *Client) AuthCheck(conf *AuthConfiguration) error {
 	if conf == nil {
 		return fmt.Errorf("conf is nil")
 	}
-	body, statusCode, err := c.do("POST", "/auth", doOptions{data: conf})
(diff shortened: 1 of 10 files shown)


fsouza:
That's a nice improvement, thanks!

fsouza:
@chuyeow FTR, I've just reverted this pull request, because of a bug reported in #378. I've opened #379 for properly handling what you reported and fixed in this PR.

Thank you!

Issue #378, opened 2015-09-11
Large responses over Unix domain socket fail with 'use of closed network connection'


awh:
After the following commit

edc2bd38d8fa79581371ea78b3c09935877a1ade

commands which yield large (> ~4k) HTTP responses over a Unix domain connection fail with the message 'use of closed network connection'. The problem stems from the deferred `dial.Close()` statement:

https://github.com/fsouza/go-dockerclient/blob/edc2bd38d8fa79581371ea78b3c09935877a1ade/client.go#L405

As long as the headers and response body fit within 4k they will be read into memory during the call to `http.ReadResponse()` before the close operation occurs. When this is not the case, JSON decode fails later on trying to read the remainder of the body from the (now closed) socket.

We noticed this in weaveworks/weave#1410; inspect of certain containers (in particular those with volume mounts) with go-dockerclient failed as the returned JSON was big enough to exceed the buffer size of the initial read.

Unfortunately removing the `defer dial.Close()`, whilst fixing the immediate problem, leaks file descriptors - the `defer resp.Body.Close()` used by the clients of `do` does not appear to close the domain socket...

fsouza:
@awh thanks for reporting it! I've just reverted edc2bd38d8fa79581371ea78b3c09935877a1ade so we can get to a proper solution for both issues.

awh:
No problem - thanks for the prompt response!

fsouza:
Closing in favor of #379.

PR #382, opened 2015-09-17
Don't read HTTP entire response body unnecessarily, take 2

### LINKED ISSUE #379: Client.do always read entire response body, which might waste memory
This has been reported and fixed in the PR #375, but as reported in #378, the fix introduced another bug: because of the ``defer dial.Close()`` call, only the first 4k bytes were available in the body, and then reading the body fails with "use of closed network connection".

We should come with a solution where Client.do don't read entire body and we don't leak file descriptors (maybe handling the socket closing in each caller of Client.do).

chuyeow:
Fixes #379, follow up to the reverted #375.

`Client.do()` now does not read the entire HTTP response from the Docker remote API, since that is an unnecessary memory / allocation burden. Instead, it returns the `*http.Response` and relies on callers to read / copy from / close the body as necessary.

When connecting to Docker via UNIX domain socket, we use a fake `http.Client` that dials to the socket, thus fixing the issue reported in #378 where the underlying network connection was prematurely closed when the HTTP response (+headers) does not fit into the default 4096-byte buffer.

### DIFF
--- a/client.go
+++ b/client.go
@@ -138,6 +138,7 @@ type Client struct {
 	requestedAPIVersion APIVersion
 	serverAPIVersion    APIVersion
 	expectedAPIVersion  APIVersion
+	unixHTTPClient      *http.Client
 }
 
 // NewClient returns a Client instance ready for communication with the given
@@ -341,6 +342,7 @@ func (c *Client) Ping() error {
 	if resp.StatusCode != http.StatusOK {
 		return newError(resp)
 	}
+	resp.Body.Close()
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


client.go
```diff
@@ -669,6 +657,41 @@ func (c *Client) getURL(path string) string {
...
+func (c *Client) unixClient() *http.Client {
+	if c.unixHTTPClient != nil {
+		return c.unixHTTPClient
+	}
+
+	socketPath := c.endpointURL.Path
+	c.unixHTTPClient = &http.Client{
+		Transport: &http.Transport{
+			Dial: func(network, addr string) (net.Conn, error) {
+				return net.Dial("unix", socketPath)
```
fsouza: Can you use c.Dialer.Dial instead of net.Dial here?

Dialer has been added to support certain customization, and we don't wanna bypass it here.

Just this change and I believe it's good to merge. Thank you very much for working on this!

chuyeow: Ah you're right, I missed that. Thanks for pointing it out - fixed.

fsouza:
Thank you!

PR #383, opened 2015-09-18
Add function SearchImagesEx to transfer authentication when search.

hopkings2008:
When calling search, there is no authentication information sent to hub.
Like docker client below:
1. docker login -u tester -p tester -e tester@test.com www.hub.com
2. docker search
docker client will send the authentication to hub also. 
But when we use the client which has login info to call SearchImages api, there is no auth header in the  http request. 
So, i add this function to transfer the authentication parameter to hub.

### DIFF
--- a/client.go
+++ b/client.go
@@ -375,6 +375,7 @@ func (c *Client) getServerAPIVersionString() (version string, err error) {
 type doOptions struct {
 	data      interface{}
 	forceJSON bool
+	headers   map[string]string
 }
 
 func (c *Client) do(method, path string, doOptions doOptions) (*http.Response, error) {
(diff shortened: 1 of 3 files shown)


hopkings2008:
@fsouza , could you please check this one? thanks a lot.

fsouza:
Hi @hopkings2008, thanks for contributing and sorry for the late response.

The search endpoint doesn't stream the response, which means that we can use the do method, extending the doOptions struct with the headers field, just like @drewwells did on #390.

hopkings2008:
@fsouza , thanks for your suggestion. I have amended my changes according to your suggestion, please help to review it again, thank.

hopkings2008:
@fsouza ,  the commit for my new change for this problem is '66c30b9960b1da2020e3c8ca34ea0b8ea729cdee'

fsouza:
@hopkings2008 thank you!

PR #388, opened 2015-09-30
Add support for ulimits during build

adrienkohlbecker:
Hello,

This adds support for the `ulimits` setting in `BuildImage()`
This parameter, while undocumented in the Remote API, is used by docker itself [here](https://github.com/docker/docker/blob/master/api/client/build.go#L261) to support the `--ulimits` argument to `docker build`.

I was not able to use `queryString()` for this as docker requires a JSON-encoded array and `queryString()` flattens the arrays.

Let me know what you think
Best

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,7 @@
 # This is the official list of go-dockerclient authors for copyright purposes.
 
 Adam Bell-Hanssen <adamb@aller.no>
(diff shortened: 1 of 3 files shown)


fsouza:
Too bad we can't extend queryString for that, as ExportImages relies on the current behavior :-(

If the JSON format gets more common in the future, we can make the ExportImages the exception case.

Thanks for contributing!

PR #390, opened 2015-10-01
add support for searching private registries fixes #389

### LINKED ISSUE #389: Add support for image search on private registries
Currently, it is not possible to pass `AuthConfiguration` for an image search. By adding auth, it would be possible to search a private registry for images.

drewwells:
if found, send headers on GET requests

Change the SearchImages method to require authconfiguration. By passing
empty auth, it will still be able to search public images.

Searching works like this:
/image/search?term=nginx (Look for nginx in index.docker.io)
/image/search?term=quay.io/nginx (look for nginx in quay.io)

It is not possible to search private images on Docker Hub yet, but at
least this will ready the go-dockerclient when that is added!

Signed-off-by: Drew Wells <drew.wells00@gmail.com>

### DIFF
--- a/client.go
+++ b/client.go
@@ -375,6 +375,7 @@ func (c *Client) getServerAPIVersionString() (version string, err error) {
 type doOptions struct {
 	data      interface{}
 	forceJSON bool
+	headers   map[string]string
 }
 
 func (c *Client) do(method, path string, doOptions doOptions) (*http.Response, error) {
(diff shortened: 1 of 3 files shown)


drewwells:
Open PR before we can merge this: https://github.com/fsouza/go-dockerclient/pull/390

fsouza:
Hi @drewwells, thanks for working on that! Could you add a check in the test to ensure that the proper registry authentication header is present?

fsouza:
Actually, can we drop this one in favor of #383, so we don't break any existing code?

I know that adding an "Ex" extension to a method/function sounds like Windows API designing, but it ensures that we don't break any existing code.

Thanks for contributing!

PR #393, opened 2015-10-08
add error judgement to prevent exception.

hopkings2008:
In client member function TagImage, add check err != nil before resp.Body.Close() is called, this will prevent the exception when err happens.


hopkings2008:
@fsouza , please help to review this change, this will prevent exception when error happens during tagimage. Thanks

fsouza:
Thank you!

PR #394, opened 2015-10-08
Export getDefaultDockerHost

cbednarski:
- Exported `getDefaultDockerHost()` so it can be used outside the library
- Inverted awkward `if` condition


fsouza:
Hi @cbednarski, thanks for contributing!

I think that it's better if we just drop the "get" and name the function DefaultDockerHost. What do you think?

cbednarski:
Sure, I will update the PR

fsouza:
Thank you!

Issue #374, opened 2015-09-08
client: trouble exiting from attached container


eliasgs:
Hi @fsouza,

Since somewhere around https://github.com/fsouza/go-dockerclient/commit/9702534da348f2aeafce8010d28a87d400cb724e, I have been having some trouble exiting from attached containers. Exiting waits for stdin and fails with: `read dev/stdin: bad file descriptor`

https://github.com/eliasgs/go-dockerclient/commit/97c85625bc5075fed8bb6ce16cebdc5e346c52b2 seems to fix it, but I am not entirely sure if there would be any undesired side effects with this?

Thanks,
Elias

fsouza:
Hi @eliasgs, can you share a reproducer? Thanks!

eliasgs:
Hi @fsouza 
Assuming you have the busybox image available, here is an example that reproduces the issue: https://github.com/eliasgs/reproducer/blob/master/main.go

Thanks:)

fsouza:
Closed via #396.

PR #396, opened 2015-10-15
Don't keep errChanIn hanging when closing

eliasgs:
Exiting an attached container waited for stdin and failed with:
"read dev/stdin: bad file descriptor".

See also issue #374

### DIFF
--- a/client.go
+++ b/client.go
@@ -623,6 +623,7 @@ func (c *Client) hijack(method, path string, hijackOptions hijackOptions) error 
 		defer func() {
 			if hijackOptions.in != nil {
 				if closer, ok := hijackOptions.in.(io.Closer); ok {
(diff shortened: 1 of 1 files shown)


fsouza:
@eliasgs maybe closing the channel is more idiomatic?

eliasgs:
@fsouza I agree, but can we be sure https://github.com/eliasgs/go-dockerclient/blob/master/client.go#L644 happens before the close?

fsouza:
@eliasgs no, we can't. Good catch, thanks.

Does that fix #374?

eliasgs:
@fsouza yes it does.

fsouza:
@eliasgs thanks!

eliasgs:
@fsouza you are welcome, and thank you!

PR #399, opened 2015-10-16
Remove use of http.DefaultClient

jefferai:
http.DefaultClient in libraries can provide dangerous and unexpected
behavior. At HC in our projects we've seen race conditions, strange TLS
behavior, and more caused from various libraries making custom changes
after sourcing http.DefaultClient, affecting all other users in the
process.

Sharing state should be explicit; use an empty client and require
downstream users to explicitly set the client to a shared one if they
need that behavior.


fsouza:
Good catch, thank you!

PR #402, opened 2015-10-20
Always send the nil signal on errChanIn when out ends

lyondhill:
If the Docker server closes the out pipe and the hijackOptions.in sent to hijack was not a io.Closer then hijack hangs indefinitely.


fsouza:
Thanks!

PR #404, opened 2015-10-22
Use cleanhttp instead of http.Client

jefferai:
Unfortunately my previous commit overlooked that even in a fresh
http.Client{}, a shared DefaultTransport is used. We made a "cleanhttp"
library that makes it easy to get a clean client and/or transport that
still use the same default values as the Go standard library, so this
commit modifies my earlier changes to use that instead.

### DIFF
--- a/client.go
+++ b/client.go
@@ -32,6 +32,7 @@ import (
 	"github.com/fsouza/go-dockerclient/external/github.com/docker/docker/opts"
 	"github.com/fsouza/go-dockerclient/external/github.com/docker/docker/pkg/homedir"
 	"github.com/fsouza/go-dockerclient/external/github.com/docker/docker/pkg/stdcopy"
+	"github.com/hashicorp/go-cleanhttp"
 )
 
 const userAgent = "go-dockerclient"
@@ -191,7 +192,7 @@ func NewVersionedClient(endpoint string, apiVersionString string) (*Client, erro
 		}
 	}
 	return &Client{
-		HTTPClient:          &http.Client{},
+		HTTPClient:          cleanhttp.DefaultClient(),
 		Dialer:              &net.Dialer{},
 		endpoint:            endpoint,
 		endpointURL:         u,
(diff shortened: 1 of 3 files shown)


peter-edge:
Really small nit, really really nitpicky nit: I feel like there's been 2000 pull requests around http.Client, can we have a function in misc.go that's just

```
func NewHTTPClient() *http.Client {
  return cleanhttp.DefaultClient()
}
```

So the next time this changes, it's just in one place?

fsouza:
Hi @jefferai, thanks for contributing again! I like @peter-edge's suggestion, could you do it? Thanks!

Also, the build failures come from the fact that the library is not vendored, could you vendor it?

jefferai:
@fsouza I'm a bit confused by the suggestion. This pull request currently takes the logic for creating a  client and uses functions from `cleanhttp`. This already implements the "next time it changes, it's in one place" request -- the changes would happen in `cleanhttp` and not affect go-dockerclient code (other than perhaps a vendoring update).

I could create `NewHTTPClient` and `NewHTTPTransport` in `misc.go` -- note that both are necessary -- and chain them to the functions from the `cleanhttp` package, but if you trust the (super super simple) logic in those functions, it seems that this pull request simply saves a step.

Alternately, if you don't trust what `cleanhttp` does, or that it will remain doing the right thing in the future, it seems like the right thing for `go-dockerclient` in that case is just to wholesale copy the logic from `cleanhttp` into `misc.go`.

fsouza:
@jefferai that's a fair point, thanks for bringing it. The trusting is not an issue, as we can go forward with the vendoring strategy. Let's go with the changes in your pull request. I will merge it and handle the vendoring.

Thank you for contributing!

jefferai:
No problem. Sorry that this came in two stages, but after I'd done the initial PR someone pointed out the shared DefaultTransport and Round 2 had to happen.

peter-edge:
Ya it's all good for the record, it was a super nit :)

PR #408, opened 2015-10-30
Fix stdin-only attach

ncdc:
Fix stdin-only attach so it doesn't wait for stdout/stderr to complete
before returning. If hijackOptions.stdout and hijackOptions.stderr are
nil, don't try to copy from the container's streams to ioutil.Discard,
as that keeps the attach session open as long as the container is
running. Instead, only copy stdout/stderr if they're non-nil.

### DIFF
--- a/client.go
+++ b/client.go
@@ -569,12 +569,6 @@ func (c *Client) hijack(method, path string, hijackOptions hijackOptions) error 
 		}
 		params = bytes.NewBuffer(buf)
 	}
-	if hijackOptions.stdout == nil {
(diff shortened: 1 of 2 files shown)


ncdc:
ping @fsouza :smile:

fsouza:
Hi @ncdc, thanks for contributing!

It's interesting that I haven't reached this bug before, as we do stdin-only attaches in tsuru. Maybe I'm just wrong and we're not doing stdin-only attaches x)

PR #411, opened 2015-11-06
Added support for filtering ImageList by a single specified name

brianhama:


fsouza:
Sorry for the late response, and thanks for contributing!

PR #412, opened 2015-11-12
Fixed port mapping for container inspect API

kobolog:
In container inspect api, if port mapping is private, i.e. not exported to the host system, then the code would mistakenly describe this port as public, and set its private counterpart to 0.


fsouza:
Hi @kobolog, thanks for contributing!

Doesn't that break existing code?

fsouza:
Just ran some tests pointing both to Docker 1.8 and Docker 1.9, and everything worked nicely. I will go forward and merge this pull request. Thanks for contributing!

PR #417, opened 2015-11-18
Extend NetworkSettings to include Networks introduced in Docker 1.9

jonuwz:
DEPRECATE following container NetworkSettings fields in API v1.21:
EndpointID, Gateway, GlobalIPv6Address, GlobalIPv6PrefixLen, IPAddress, IPPrefixLen, IPv6Gateway and MacAddress.
Those are now specific to the bridge network. Use NetworkSettings.Networks to inspect the networking settings of a container per network

### DIFF
--- a/container.go
+++ b/container.go
@@ -125,25 +125,38 @@ type PortBinding struct {
 // and its value as found in NetworkSettings should always be nil
 type PortMapping map[string]string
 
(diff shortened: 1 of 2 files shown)


fsouza:
@jonuwz thanks for contributing! I was wondering whether we should document that, but we may add this in the future. Merging.

PR #419, opened 2015-11-20
If there is no scheme in the endpoint, use tcp, to match docker cli behaviour

paulbellamy:


fsouza:
Nice change, thank you!

PR #421, opened 2015-11-20
Fix volume creation API

dhiltgen:
When volume creation transitioned from experimental to
supported in 1.9, the API changed slightly.
Reference: https://goo.gl/pBUbZ9

Signed-off-by: Daniel Hiltgen <daniel.hiltgen@docker.com>


fsouza:
Thank you Daniel!

PR #428, opened 2015-12-01
Add IPAM options and config to CreateNetworkOptions struct

bboreham:
These were added in Docker 1.9 (ref #407)


fsouza:
Thank you!

Issue #426, opened 2015-11-30
Network api links are wrong


bboreham:
There are various links in `network.go` to https://goo.gl/FDkCdQ or https://goo.gl/4hCNtZ; this is the driver side of the API, not the client side that this library should be referring to.

Most of the structures are in https://github.com/docker/docker/blob/master/api/types/types.go, but this file does not explain everything linked to so a straight search-and-replace falls short.

PR #429, opened 2015-12-01
Fix network structs for Docker 1.9

bboreham:
Resolves part of #407 and also #426.

Perhaps the previous code was working off a pre-release version of the API?

Note in most cases I have gone with the current Docker field names, e.g. `Type` -> `Driver`, `Endpoints` -> `Containers`, which will break people's code, but that code couldn't have worked very well anyway against Docker 1.9.

### DIFF
--- a/network.go
+++ b/network.go
@@ -17,21 +17,25 @@ var ErrNetworkAlreadyExists = errors.New("network already exists")
 
 // Network represents a network.
 //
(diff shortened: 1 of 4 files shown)


fsouza:
Indeed the previous code was for the experimental release of the networking API. Thanks for contributing again!

PR #431, opened 2015-12-06
Added RepoDigests as a field of the Image struct

lioryanko:
For reference: https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#inspect-an-image

Thanks!


fsouza:
@lioryanko thanks!

PR #432, opened 2015-12-07
Add cancellable attach.

### LINKED ISSUE #161: Allow cancellation of AttachToContainer and Logs requests
Currently, I found no way to stop `AttachToContainer` or `Logs` (with Follow:true) requests from the client.  It would be great to have a way to close these connections without quitting the client-side process.

tomwilkie:
Let me know what you think to the approach - I can extend it to exec also.

Fixes #161

### DIFF
--- a/container.go
+++ b/container.go
@@ -1040,6 +1040,28 @@ func (c *Client) AttachToContainer(opts AttachToContainerOptions) error {
 		return &NoSuchContainer{ID: opts.Container}
 	}
 	path := "/containers/" + opts.Container + "/attach?" + queryString(opts)
+	cw, err := c.hijack("POST", path, hijackOptions{
+		success:        opts.Success,
+		setRawTerminal: opts.RawTerminal,
+		in:             opts.InputStream,
+		stdout:         opts.OutputStream,
+		stderr:         opts.ErrorStream,
+	})
+	if err != nil {
+		return err
+	}
+	return cw.Wait()
+}
+
+// AttachToContainerNonBlocking attaches to a container, using the given options.
+// This function does not block.
+//
+// See https://goo.gl/NKpkFk for more details.
+func (c *Client) AttachToContainerNonBlocking(opts AttachToContainerOptions) (CloseWaiter, error) {
+	if opts.Container == "" {
+		return nil, &NoSuchContainer{ID: opts.Container}
+	}
+	path := "/containers/" + opts.Container + "/attach?" + queryString(opts)
 	return c.hijack("POST", path, hijackOptions{
 		success:        opts.Success,
 		setRawTerminal: opts.RawTerminal,
--- a/container.go
+++ b/container.go
@@ -1036,17 +1036,7 @@ type AttachToContainerOptions struct {
 //
 // See https://goo.gl/NKpkFk for more details.
 func (c *Client) AttachToContainer(opts AttachToContainerOptions) error {
-	if opts.Container == "" {
-		return &NoSuchContainer{ID: opts.Container}
-	}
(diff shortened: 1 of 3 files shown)


tomwilkie:
I've extended this to StartExec now.  What do you think?  Would you prefer we pass in a channel the user can close instead?

tomwilkie:
@fsouza ping?

fsouza:
Hi @tomwilkie, thank you very much for working on that! I'm planning to have a look at it later today or tomorrow, in the worst case. I've seen your comment on #161, the idea looks promising.

Sorry for taking that long to get back to you, and thank you again for contributing!

tomwilkie:
Not a problem!  Let me know which way you prefer (returning an WaitCloser vs passing a channel) and I'll code it up.

fsouza:
@tomwilkie I do prefer the WaitCloser approach, looking at the code at this moment!

container.go
```diff
@@ -1040,6 +1040,28 @@ func (c *Client) AttachToContainer(opts AttachToContainerOptions) error {
...
+		setRawTerminal: opts.RawTerminal,
+		in:             opts.InputStream,
+		stdout:         opts.OutputStream,
+		stderr:         opts.ErrorStream,
+	})
+	if err != nil {
+		return err
+	}
+	return cw.Wait()
+}
```
fsouza: Maybe AttachToContainer could just call AttachToContainerNonBlocking and then wait on the CloseWaiter? It would avoid code duplication.

The same applies for StartExec.

fsouza:
I liked your implementation, I was wondering  whether we could make CloseWaiter a parameter in AttachToContainerOptions (and StartExecOptions), along with a boolean flag for identifiying whether we want to block or not, but maybe just having another method is a better idea!

May I invoke others here? @peter-edge @cezarsa @andrewsmedina what do you all think? :-)

container.go
tomwilkie: Yeah, good idea.  Will do.

tomwilkie:
> I liked your implementation,

Thanks!

> I was wondering whether we could make CloseWaiter a parameter in AttachToContainerOptions (and StartExecOptions), along with a boolean flag for identifiying whether we want to block or not, but maybe just having another method is a better idea!

We could add it as a "out" parameter in the struct I guess - and that would solve the issue of naming the second copy of the function.  Let me have a look and see how it feels - my initial thought is having the method in question modify the struct being passed in (and/or change its basic behaviour based on a bool) isn't particularly desirable.

tomwilkie:
> May I invoke others here? @peter-edge @cezarsa @andrewsmedina what do you all think? :-)

Oh yes please!

exec.go
```diff
@@ -83,8 +83,24 @@ type StartExecOptions struct {
 //
 // See https://goo.gl/iQCnto for more details
 func (c *Client) StartExec(id string, opts StartExecOptions) error {
+	cw, err := c.StartExecNonBlocking(id, opts)
+	if err != nil {
+		return err
+	}
+	if cw != nil {
```
fsouza: Is there a scenario where both cw and err are nil?

fsouza: Sorry, just checked and there's hahaha

Sorry about that.

fsouza:
I've spoken with @andrewsmedina and @cezarsa offline and we believe that it's good to merge. Thanks for contributing!

tomwilkie:
Thanks!

pwaller:
Just saw this one fly by. A bit unfortunate it doesn't use x/net/context. Is it straightforward to build context-based cancellation on top of this?

tomwilkie:
@pwaller yes it should be relatively straight forward to hook this up to context-based cancellation.

PR #437, opened 2015-12-17
Added volumes api to the test server.

chosenken:
Hi.  I'm using the go-dockerclient library for a project, and needed support for volumes in my unit tests.  This pull request covers adding the volume APIs for docker based on the remote 1.21 api to the test server.  So far it has been working for me in emulating my needs for volumes, which mainly pertains to the create/deletion of volumes.

Added support for the following volume API's:

List Volumes: GET /volumes
Create Volume: POST /volumes/create
Inspect Volume: GTET /volumes/(name)
Remove Volume: DELETE /volumes/(name)

### DIFF
--- a/testing/server.go
+++ b/testing/server.go
@@ -1133,11 +1133,7 @@ func (s *DockerServer) listVolumes(w http.ResponseWriter, r *http.Request) {
 	s.volMut.RLock()
 	result := make([]docker.Volume, 0, len(s.volStore))
 	for _, volumeCounter := range s.volStore {
-		result = append(result, docker.Volume{
-			Name:       volumeCounter.volume.Name,
-			Driver:     volumeCounter.volume.Driver,
-			Mountpoint: volumeCounter.volume.Mountpoint,
-		})
+		result = append(result, volumeCounter.volume)
 	}
 	s.volMut.RUnlock()
 	w.Header().Set("Content-Type", "application/json")
@@ -1172,14 +1168,9 @@ func (s *DockerServer) createVolume(w http.ResponseWriter, r *http.Request) {
 
 	// If the volume already exists, don't re-add it.
 	exists := false
-	s.volMut.RLock()
+	s.volMut.Lock()
 	if s.volStore != nil {
-		for _, vol := range s.volStore {
-			if vol.volume.Name == volume.Name {
-				exists = true
-				break
-			}
-		}
+		_, exists = s.volStore[volume.Name]
 	} else {
 		// No volumes, create volStore
 		s.volStore = make(map[string]*volumeCounter)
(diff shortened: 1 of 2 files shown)


testing/server.go
```diff
@@ -1117,3 +1128,111 @@ func (s *DockerServer) createNetwork(w http.ResponseWriter, r *http.Request) {
...
+		for _, vol := range s.volStore {
+			if vol.volume.Name == volume.Name {
+				exists = true
+				break
+			}
+		}
+	} else {
+		// No volumes, create volStore
+		s.volStore = make(map[string]*volumeCounter)
+	}
```
fsouza: Maybe here you should call s.volMut.RUnlock(), and then inside the if condition do the following:

```go
if !exists {
    s.volMut.Lock()
    defer s.volMut.Unlock()
    if _, ok := s.volStore[volume.Name]; !ok {
        s.volStore[volume.Name] = &volumeCounter{volume: *volume, count: 0}
    }
}
```

You may alternatively just use a regular lock, as this is a server intended for testing purposes.

testing/server.go
```diff
@@ -1117,3 +1128,111 @@ func (s *DockerServer) createNetwork(w http.ResponseWriter, r *http.Request) {
...
+
+	// If the volume already exists, don't re-add it.
+	exists := false
+	s.volMut.RLock()
+	if s.volStore != nil {
+		for _, vol := range s.volStore {
+			if vol.volume.Name == volume.Name {
+				exists = true
+				break
+			}
```
fsouza: You can replace this for with:

```go
_, exists = s.volStore[volume.Name]
```

testing/server.go
```diff
@@ -1117,3 +1128,111 @@ func (s *DockerServer) createNetwork(w http.ResponseWriter, r *http.Request) {
...
+func (s *DockerServer) findVolume(name string) (*volumeCounter, error) {
+	s.volMut.RLock()
+	defer s.volMut.RUnlock()
+	if s.volStore != nil {
+		for _, vol := range s.volStore {
+			if vol.volume.Name == name {
+				return vol, nil
+			}
+		}
+	}
```
fsouza: You can replace this if and along with the for with the following code:

```go
vol, ok := s.volStore[name]
if !ok {
    return nil, errors.New("no such volume")
}
return vol, nil
```

fsouza:
Hi @chosenken, thanks for contributing! The change looks interesting, but there are some improvements:

- you can use comma-ok to check for presence in a map (``val, ok := map[key]``)
- when using a RWMutex, RLock is intended for reads and Locks for writes, so a common idiom is:

```go
mutex.RLock()
// check that the value doesn't exist in the map
mutex.RUnlock()
mutex.Lock()
// check that the value haven't changed
// modify the value if it haven't changed
mutex.Unlock()
```

The problem is that it's too complex for a testing server, so you could simply use a regular mutex, and do the following:

```go
mutex.Lock()
defer mutex.Unlock()
// check that the value doesn't exist in the map
// change the map
```

What do you think of these changes?

chosenken:
They sound good!  I'm still getting into go, so these comments help a lot.  I'll work on them, thanks!

fsouza:
@chosenken no problem. Please let me know if you need further help!

testing/server.go
```diff
@@ -1117,3 +1128,99 @@ func (s *DockerServer) createNetwork(w http.ResponseWriter, r *http.Request) {
...
+			count:  0,
+		}
+	}
+	s.volMut.Unlock()
+	w.WriteHeader(http.StatusCreated)
+	json.NewEncoder(w).Encode(volume)
+}
+
+func (s *DockerServer) inspectVolume(w http.ResponseWriter, r *http.Request) {
+	s.volMut.Lock()
```
fsouza: Since you're just reading the volume here, you could use RLock. What do you think?

Otherwise, LGTM. Thanks for contributing!

fsouza:
I've merged it and did the adjust myself, thank you very much for contributing!

PR #439, opened 2015-12-23
Change querystring key for Message

hatchan:
According to the documentation the correct key is `comment`, not `m`: https://docs.docker.com/engine/reference/api/docker_remote_api_v1.20/#create-a-new-image-from-a-container-s-changes


fsouza:
Hi @hatchan, thanks for contributing. IIRC, it used to be m, but at some point they changed it. Merging now!

PR #444, opened 2016-01-15
testing/server: add archive route to upload a file to container.

tarsisazevedo:
add missing route to upload a file to container.

### DIFF
--- a/testing/server.go
+++ b/testing/server.go
@@ -511,6 +511,7 @@ func (s *DockerServer) uploadToContainer(w http.ResponseWriter, r *http.Request)
 	container, _, err := s.findContainer(id)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusNotFound)
+		return
 	}
 	if !container.State.Running {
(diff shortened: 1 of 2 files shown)


testing/server.go
```diff
@@ -503,6 +506,22 @@ func (s *DockerServer) statsContainer(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *DockerServer) uploadToContainer(w http.ResponseWriter, r *http.Request) {
+	id := mux.Vars(r)["id"]
+	container, _, err := s.findContainer(id)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
```
fsouza: You need to add an empty return here.

fsouza: Could you also please add a test for this case. I know that this test is missing in other handlers, but please add it here x)

Thank you!

tarsisazevedo: done! :)

fsouza:
I'll go ahead and merge it, I believe that the build will be green.

Thanks for contributing!

PR #446, opened 2016-01-19
Add Docker Network 'IPAM' field

bboreham:
I missed this from #428 

Also the URL for info has changed again, as Docker have moved that file.


fsouza:
Thank you!

PR #451, opened 2016-01-22
Allow -distro-version to append to APIVersion

timothysc:
In testing on fedora+rhel errors are returned on what are valid version appenders for distributions.  

1.10.0-rc1-fc24

This is a quick fix to allow.


fsouza:
Hey @timothysc, thanks for contributing! Can you add a test for this case? An entry in the following test should be enough: https://github.com/fsouza/go-dockerclient/blob/master/client_test.go#L345. Does that sound reasonable?

Thanks for contributing!

fsouza:
I will add the case myself, thanks for contributing! :-)

PR #453, opened 2016-01-26
Add missing swap memory stats fields to Stats struct

cezarsa:
These fields are not currently listed at https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#get-container-stats-based-on-resource-usage but they are exposed in cgroup memory stats and docker send them on /stats response.


fsouza:
LGTM, please feel free to merge :D

PR #472, opened 2016-02-22
fix buildarg serialization

justone:
When using go-dockerclient on docker 1.10.1, I encountered this error:

```
2016/02/22 13:39:48 API error (500): json: cannot unmarshal array into Go value of type map[string]string
```

I tracked it down to the serialization of the BuildArg structs.  I switched it out with a regular string map and it works now.

Previous: `[{"Name":"SOME_VAR","Value":"some_value"}]`
New: `{"SOME_VAR":"some_value"}`

### DIFF
--- a/image.go
+++ b/image.go
@@ -465,7 +465,11 @@ func (c *Client) BuildImage(opts BuildImageOptions) error {
 	}
 
 	if len(opts.BuildArgs) > 0 {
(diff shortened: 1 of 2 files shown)


Jbbouille:
+1

fsouza:
I was looking into Docker code to see if this changed, but looks like it was a bad review of mine, I apologize for that, and thanks for fixing it!

PR #475, opened 2016-03-02
add missing ContainerNetwork.NetworkID field

codekitchen:
This field is documented in the response at
https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#inspect-a-container


fsouza:
Thank you!

Issue #474, opened 2016-03-02
Client.Info() return does not correctly reflect returned data


cezarsa:
Currently [(*Client).Info()](https://github.com/fsouza/go-dockerclient/blob/2705b71/misc.go#L28) return is an `Env` instance which is in fact a `[]string` where each entry is in the format `key=value`.

This does not correctly represent the response from `GET /info` route. According to [docker api docs](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#display-system-wide-information) the response is a complex json. My opinion is that go-dockerclient should have a custom struct mapping these fields.

fsouza:
Closing, as #476 has been merged.

PR #476, opened 2016-03-05
Add some Network, Container and Misc changes from the 1.22 API

nickvanw:
This PR adds the following capabilities from the 1.22 API:

* POST /container/(name)/update updates the resources of a container.
* GET /containers/json now returns the list of networks of containers.
* GET /info Now returns Architecture and OSType fields, providing information about the host architecture and operating system type that the daemon runs on.
* POST /containers/create now allows you to set a read/write rate limit for a device (in bytes per second or IO per second).
* GET /networks now supports filtering by name, id and type.
* GET /info now includes the number of containers running, stopped, and paused.
* POST /networks/create now supports restricting external access to the network by setting the internal field.
* POST /networks/(id)/disconnect now includes a Force option to forcefully disconnect a container from network
* GET /info can now return a SystemStatus field useful for returning additional information about applications that are built on top of engine.

Most of the changes are completely backward compatible - the biggest breaking change is that the info route now returns a proper struct. If this change is too breaking, let me know and I can create a new method returning this info (or something else!)

I tried to match the style of the other endpoints and tests, let me know if anything looks awry! 

I'd love to get the Event changes in, as well as some of the IP/IPAM changes. Those are next on my list.

Should address #474

### DIFF
--- a/container.go
+++ b/container.go
@@ -53,7 +53,7 @@ type APIContainers struct {
 	SizeRootFs int64             `json:"SizeRootFs,omitempty" yaml:"SizeRootFs,omitempty"`
 	Names      []string          `json:"Names,omitempty" yaml:"Names,omitempty"`
 	Labels     map[string]string `json:"Labels,omitempty" yaml:"Labels,omitempty"`
-	Networks   NetworkList       `json:"NetworkSettings,omitempty" yaml:"NetworkSettings,omitempty`
+	Networks   NetworkList       `json:"NetworkSettings,omitempty" yaml:"NetworkSettings,omitempty"`
 }
 
 type NetworkList struct {
@@ -523,7 +523,7 @@ type BlockWeight struct {
 // See https://goo.gl/FSdP0H for more details.
 type BlockLimit struct {
 	Path string `json:"Path,omitempty"`
-	Rate string `json:"Rate",omitempty`
+	Rate string `json:"Rate,omitempty"`
 }
 
 // HostConfig contains the container options related to starting a container on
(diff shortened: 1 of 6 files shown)


nickvanw:
Looks like some sporadic build failures due to `vet` and other things - I'll fix this up soon and add more tests around some of the new parameters.

fsouza:
Wow, awesome work! Sorry for taking two days to get back to you, I will have a look in the code while you work on the fix for vet at tip.

container.go
```diff
@@ -52,7 +52,12 @@ type APIContainers struct {
 	SizeRw     int64             `json:"SizeRw,omitempty" yaml:"SizeRw,omitempty"`
 	SizeRootFs int64             `json:"SizeRootFs,omitempty" yaml:"SizeRootFs,omitempty"`
 	Names      []string          `json:"Names,omitempty" yaml:"Names,omitempty"`
-	Labels     map[string]string `json:"Labels,omitempty" yaml:"Labels, omitempty"`
+	Labels     map[string]string `json:"Labels,omitempty" yaml:"Labels,omitempty"`
+	Networks   NetworkList       `json:"NetworkSettings,omitempty" yaml:"NetworkSettings,omitempty`
```
fsouza: vet pointed that, but I'd like to comment as well: there a quote missing between the last omitempty and `

fsouza:
The code looks good, just fixing the vet errors and it should be good to merge. I'm not too worried about the Info change. Ideally we would not change it, but it's a big change in the API and people can easily update their code, for better. I see like Info has become more useful now, so it's more likely that people want the new version.

Thanks for working on that!

nickvanw:
@fsouza looks like the only test failures are `vet` issues in `tip` only due to `server.go` and `server_test.go` -- I can probably fix those up at some point but this week, but they probably don't need to block this PR.

fsouza:
Cool, thank you! I'll fix the vet warning myself.

Issue #471, opened 2016-02-21
1.22 Events Format Change


schmatz:
The events fields have changed from [1.21](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#monitor-docker-s-events) to [1.22](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#monitor-docker-s-events). Are there plans to change the `APIEvents` fields accordingly?

fsouza:
Yes, but we should keep go-dockerclient compatible with older versions of the API as well.

nickvanw:
Hey @fsouza! I'm looking at picking up changing the `APIEvents` over to use the new schema in 1.22, but I'm having a hard time picturing how to make sure it's backwards compatible with 1.21 and prior versions. 

Because the Docker API change involves adding additional events, as well as radically changing the schema, trying to shoehorn information into a backwards compatible struct feels like it would only cause confusion and create a lot of cruft in the `go-dockerclient` API. 

Any opinions/ideas? cc @schmatz

schmatz:
We might be able update the struct to reflect the new API but translate events from the older APIs to fit in the new struct, with pointers to fields that may not exist for older versions of the API. It might be possible to do this transparently. I haven't had the time to look into this in depth, so I defer to the judgement of @fsouza.

If there are a ton of new features that also exhibit the same backwards compatibility issues and changing the library dramatically isn't an option, might be best to make a fork.

nickvanw:
If anyone is looking for this ASAP and doesn't need backwards compatibility, https://github.com/nickvanw/go-dockerclient/commit/0dd806834d0002ea156de70ee2f59979f75fc1d7 should get you what you're looking for. I haven't extensively tested it, but it should work.

fsouza:
Closing, as #477 has been merged. @nickvanw, thanks for working on that!

PR #477, opened 2016-03-06
events: change to API 1.22 format, maintain compatibility

nickvanw:
This was a tricky one, but I think I've gotten the change completely forward and backwards compatible. 

The changes to the Docker API introduced event hooks for more than just Containers and Images, and the `Action` attribute is overridden (meaning you can 'create' networks, containers, volumes, etc). Because of this, if an event was not present before 1.22, we prepend the `Action` attribute with the `Type`, and separate them by a `:`.

This means that creating a network will have a Status of `network:create`, whereas a container create will still be just `create`, which is the way it would work with 1.21. 

There's some funny stuff that happens with `From` and `Attributes` -- I'm going to play around with the attributes for other event types to make sure it works properly.

This behavior probably deserves some documentation - let me know if you want that to live on `APIEvents` or elsewhere. 

Thoughts? Addresses #471

### DIFF
--- a/event.go
+++ b/event.go
@@ -18,12 +18,38 @@ import (
 	"time"
 )
 
-// APIEvents represents an event returned by the API.
+// APIEvents represents events coming from the Docker API
+// The fields in the Docker API changed in API version 1.22, and
+// events for more than images and containers are now fired off.
+// To maintain forward and backward compatibility, go-dockerclient
+// replicates the event in both the new and old format as faithfully as possible.
+//
+// For events that only exist in 1.22 in later, `Status` is filled in as
+// `"Type:Action"` instead of just `Action` to allow for older clients to
+// differentiate and not break if they rely on the pre-1.22 Status types.
+//
+// The transformEvent method can be consulted for more information about how
+// events are translated from new/old API formats
 type APIEvents struct {
-	Status string `json:"Status,omitempty" yaml:"Status,omitempty"`
-	ID     string `json:"ID,omitempty" yaml:"ID,omitempty"`
-	From   string `json:"From,omitempty" yaml:"From,omitempty"`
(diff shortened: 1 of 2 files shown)


fsouza:
That's a nice solution, thanks for working on that and ensuring that it's backward compatible. I agree that we should have more documentation on this behavior, in ``APIEvents``.

You mentioned "some funny stuff" happening and that you'd play around to ensure proper behavior, did you find anything odd? What were the funny stuff?

Thank you again!

nickvanw:
Everything seems to be good! I wanted to make sure that the `Attributes` map would get copied over to `From` right in new events - the keys in the map aren't documented anywhere in the Docker models, and none of the examples in the Docker API docs indicate what they'll look like. I did some testing yesterday and today, and it looks like everything is good:

```
&docker.APIEvents{Action:"connect", Type:"network", Actor:docker.APIActor{ID:"b4e6762117b33651bbfa9ad6187c02c1a732201438388b95cf497080e58afe0a", Attributes:map[string]string{"container":"9b71ac7dfb67c73007c6dc51d5a3f26d09d281a600f22809361db76da5fbdb0c", "name":"bridge", "type":"bridge"}}, Status:"network:connect", ID:"b4e6762117b33651bbfa9ad6187c02c1a732201438388b95cf497080e58afe0a", From:"", Time:1457403712, TimeNano:1457403712849271292}
```

That's an example from a standard network create that will now get fired off from a simple `docker run`. It looks like everything is good as far as other create-style events -- `From` only gets copied over in the right case. Same goes for containers, and other events. 

I'll add more documentation to the `APIEvents` struct now!

fsouza:
@nickvanw LGTM, thank you!

PR #469, opened 2016-02-17
Add in the new network connection parameters as per 1.22 api version

lyondhill:
Fixes #460

PR #480, opened 2016-03-10
network: add new network connection params for 1.22

nickvanw:
This duplicates #469 - it implements the endpoint change in 1.22:

* POST /networks/(id)/connect now allows you to set the static IPv4 and/or IPv6 address for the container.

Unfortunately, the struct accepted in both ConnectNetwork and DisconnectNetwork has grown different, but it would be a breaking change to move to using different structs in the client. Not the end of the world, but it will cause some extra bytes to get sent over the network.

If we're willing to make the breaking change, I'd gladly split the two structs up, but I figure it's best to keep compatibility here.

### DIFF
--- a/network.go
+++ b/network.go
@@ -187,10 +187,33 @@ func (c *Client) RemoveNetwork(id string) error {
 
 // NetworkConnectionOptions specify parameters to the ConnectNetwork and DisconnectNetwork function.
 //
-// See https://goo.gl/6GugX3 for more details.
+// See https://goo.gl/RV7BJU for more details.
 type NetworkConnectionOptions struct {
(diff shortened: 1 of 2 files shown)


nickvanw:
Only non-allowed failure appears to be a flappy test unrelated to these changes!

fsouza:
@nickvanw we can take it as a lesson to the future, always use a struct per Docker call. omitempty should help reducing the amount of extra bytes, and I don't think that a lot of people would suffer because of those bytes. Using a single struct was a mistake, but we have to live with it now haha

Thank you very much for working on that and all your contributions!

PR #482, opened 2016-03-11
Added State.StateString to match docker's state rendering

paulbellamy:
Docker uses `State()` to refer to the human-readable version of the state (shown in `docker ps`) and `StateString()` to refer to the machine-readable version. I've added those here to match.

### DIFF
--- a/container.go
+++ b/container.go
@@ -14,6 +14,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/fsouza/go-dockerclient/external/github.com/docker/go-units"
 )
 
 // ErrContainerAlreadyExists is the error returned by CreateContainer when the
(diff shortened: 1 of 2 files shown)


paulbellamy:
Downside is it adds a dependency on [github.com/docker/go-units](github.com/docker/go-units)

fsouza:
Wow, somehow I missed this PR. I apologize!

I believe it's fine to depend on go-units, as it contains no further third-party dependencies.

The change looks fair, but can you have a look at the build failure? Thanks for contributing!

paulbellamy:
Seems to be some sort of issue with vendoring

container.go
```diff
@@ -108,15 +112,55 @@ type State struct {
 	FinishedAt time.Time `json:"FinishedAt,omitempty" yaml:"FinishedAt,omitempty"`
 }
 
-// String returns the string representation of a state.
+// String returns a human-readable description of the state
 func (s *State) String() string {
 	if s.Running {
 		if s.Paused {
```
peter-edge: Is there never a case where `s.Paused && !s.Running`?, Similar question for `Restarting`, and the assumption that the fallthrough is `exited.`

container.go
```diff
@@ -108,15 +112,55 @@ type State struct {
...
+
+	if s.FinishedAt.IsZero() {
+		return ""
+	}
+
+	return fmt.Sprintf("Exited (%d) %s ago", s.ExitCode, units.HumanDuration(time.Now().UTC().Sub(s.FinishedAt)))
+}
+
+// StateString returns a single string to describe state
+func (s *State) StateString() string {
```
peter-edge: Same questions as before - I should come up with an alternative proposal before saying this, but this just seems prone to bugs. How does state get set? Is there a spec in the Docker API?

paulbellamy:
@peter-edge These are taken from https://github.com/docker/docker/blob/e5a3f86e447dd659da3c2e759f3c088a0bfcfe3d/container/state.go#L40, so if there is a bug in them with multiple bools being set, it's probably also a bug in the docker API.

peter-edge:
Got it - do we handle RemovalInProgress/should we?

paulbellamy:
> Got it - do we handle RemovalInProgress/should we?

Ah, that was added after I grabbed the first version. We definitely should. :)

paulbellamy:
Fixed up the tests, but it seems like there is an intermittent failure on one of the builds?

fsouza:
@paulbellamy thanks for working on that. The build failure was indeed intermittent, it's now green and good to merge.

PR #491, opened 2016-03-28
event: discover version more reliably, be more defensive with field copying

### LINKED ISSUE #490: 1.22 Event Format not working as expected
@nickvanw 

Thanks for the work on the new Docker events in 1.22, but I'm having a small issue with it. 
Events coming from API version 1.22 on my Docker 1.10.3 instance aren't showing up as expected. (I may be just misunderstanding the way it's supposed to work.)  

Here is a container create event as sent by the Docker API:

```
{
    "Action": "create",
    "Actor": {
        "Attributes": {
            "HAProxyMode": "http",
            "HealthCheck": "HttpGet",
            "HealthCheckArgs": "http://127.0.0.1:39051/status/check",
            "ServicePort_8080": "17801",
            "image": "datanerd.us/siteeng/sample-app-go:latest",
            "name": "sample-app-client-go-69818c1223ddb5"
        },
        "ID": "a925eaf4084d5c3bcf337b2abb05f566ebb94276dff34f6effb00d8ecd380e16"
    },
    "Type": "container",
    "from": "datanerd.us/siteeng/sample-app-go:latest",
    "id": "a925eaf4084d5c3bcf337b2abb05f566ebb94276dff34f6effb00d8ecd380e16",
    "status": "create",
    "time": 1459133932,
    "timeNano": 1459133932961735842
}
```

And here's the APIEvents struct returned by go-dockerclient (I did a printf("%#v")):

```
&docker.APIEvents{
        Action:"create",
        Type:"container",
        Actor:docker.APIActor{
                ID:"a925eaf4084d5c3bcf337b2abb05f566ebb94276dff34f6effb00d8ecd380e16",
                Attributes:map[string]string{
                        "image":"datanerd.us/siteeng/sample-app-go:latest"
                }
        },
        Status:"create",
        ID:"a925eaf4084d5c3bcf337b2abb05f566ebb94276dff34f6effb00d8ecd380e16",
        From:"datanerd.us/siteeng/sample-app-go:latest",
        Time:1459133932,
        TimeNano:1459133932961735842
}
```
It looks like the event is being mistakenly transformed because it's meeting the '<= 1.21' test in transformEvent:
```
func transformEvent(event *APIEvents) {
	// if <= 1.21, `status` and `ID` will be populated
	if event.Status != "" && event.ID != "" {
```

I notice that both 'status' and 'ID' are both populated in the JSON returned by the Docker API, even though the version is 1.22. 

I'm using go-dockerclient commit ID 7c07ffce0f7e14a4da49ce92a2842d4e87be1c1e, and my Docker instance looks like:

```
# docker version
Client:
 Version:      1.10.3
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   20f81dd
 Built:        Thu Mar 10 15:39:25 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.10.3
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   20f81dd
 Built:        Thu Mar 10 15:39:25 2016
 OS/Arch:      linux/amd64
```

Maybe we could get the API version with the /version endpoint, and transform based on that? What do you think?

Thanks!

nickvanw:
Unfortunately, the heuristics I used before this weren't as reliable as I wanted - for some container events, we would mistakenly think that they were <= API version 1.21, which would result in us copying over some fields that we didn't want to.

Now, we look at fields that will only exist in versions 1.22 and above. In addition, we won't overwrite some fields unless they are empty, to avoid problems in the future.

This fixes #490 (and uses the create event as a test case).

### DIFF
--- a/event.go
+++ b/event.go
@@ -334,8 +334,8 @@ func (c *Client) eventHijack(startTime int64, eventChan chan *APIEvents, errChan
 // transformEvent takes an event and determines what version it is from
 // then populates both versions of the event
 func transformEvent(event *APIEvents) {
-	// if <= 1.21, `status` and `ID` will be populated
-	if event.Status != "" && event.ID != "" {
+	// if event version is <= 1.21 there will be no Action and no Type
+	if event.Action == "" && event.Type == "" {
 		event.Action = event.Status
 		event.Actor.ID = event.ID
 		event.Actor.Attributes = map[string]string{}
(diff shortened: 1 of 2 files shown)


fsouza:
Cool, thanks for handling the issue and creating this PR!

LGTM

PR #492, opened 2016-03-28
Fixed parsing of docker client configuration.

romank87:
If file ~/docker/config.json contains something except auth information parsing will fail. But according to the [doc](https://docs.docker.com/engine/reference/commandline/cli/) config.json could contain a bunch of other information: HttpHeaders, detachKeys etc...
 This patch solves this issue.


fsouza:
LGTM, thanks!

The build failure was flaky, I'm just waiting it to become green before merging the PR.

PR #496, opened 2016-03-31
Repotags are also in docker inspect calls

samifruit514:
We need to get tags "aliases" from an inspect image api call. Docker inspect from command line returns a structure with RepoTags in it.

I would be nice if we could get from InspectImage also.


fsouza:
LGTM, thank you!

PR #497, opened 2016-04-02
Update vendored go packages for OpenBSD support

frenkel:
This will allow go-dockerclient to be used on OpenBSD. I saw a different pull request that included other changes as well, so made this simpler one.


fsouza:
Thank you!

PR #501, opened 2016-04-14
Add graphdriver support

pmorie:
@sjenning @fsouza 

cc @vishh


fsouza:
Hi @pmorie, is this a "read-only" attribute of the container? Or do we wanna include this in Config/HostConfig for launching new containers through the API as well?

fsouza:
BTW, this can come in another PR. LGTM, thanks for contributing!

PR #502, opened 2016-04-14
updates for API version 1.23

peter-edge:
Signed-off-by: Peter Edge <peter.edge@gmail.com>

### DIFF
--- a/container.go
+++ b/container.go
@@ -605,6 +605,7 @@ type HostConfig struct {
 	DNSSearch            []string               `json:"DnsSearch,omitempty" yaml:"DnsSearch,omitempty"`
 	ExtraHosts           []string               `json:"ExtraHosts,omitempty" yaml:"ExtraHosts,omitempty"`
 	VolumesFrom          []string               `json:"VolumesFrom,omitempty" yaml:"VolumesFrom,omitempty"`
+	UsernsMode           string                 `json:"UsernsMode,omitempty" yaml:"UsernsMode,omitempty"`
 	NetworkMode          string                 `json:"NetworkMode,omitempty" yaml:"NetworkMode,omitempty"`
 	IpcMode              string                 `json:"IpcMode,omitempty" yaml:"IpcMode,omitempty"`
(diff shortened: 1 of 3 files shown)


container.go
```diff
@@ -262,6 +275,7 @@ type Config struct {
 	MemorySwap        int64               `json:"MemorySwap,omitempty" yaml:"MemorySwap,omitempty"`
 	MemoryReservation int64               `json:"MemoryReservation,omitempty" yaml:"MemoryReservation,omitempty"`
 	KernelMemory      int64               `json:"KernelMemory,omitempty" yaml:"KernelMemory,omitempty"`
+	PidsLimit         int64               `json:"PidsLimit,omitempty" yaml:"PidsLimit,omitempty"`
```
jaxi: Thanks for the updating! `PidsLimit` is something I'm looking for to deal with the fork bombing. 

btw how about having some test against the `PidsLimit`?

peter-edge: What kind of test are you looking for? :)

peter-edge:
@fsouza I think travis is being flaky, also with https://github.com/fsouza/go-dockerclient/pull/504

container.go
jaxi: Maybe the [container inspecting](https://github.com/fsouza/go-dockerclient/blob/bf97c77db7c945cbcdbf09d56c6f87a66f54537b/container_test.go#L183)?

peter-edge: I want to say this might be out of scope for this PR (which is a bad way to say "I'm being lazy"), so maybe we can add it after we get this in :)

fsouza: Maybe you could add some integration test to the integration suite, but then we'd need to figure a way of running some tests against certain versions of Docker only.

fsouza: Anyway, this can be handled in another PR.

fsouza:
@peter-edge the build is now green :D 

LGTM, thanks for contributing!

container.go
jaxi: np - cheers guys!

PR #505, opened 2016-04-15
Add AuthStatus for /auth API call

peter-edge:
Signed-off-by: Peter Edge <peter.edge@gmail.com>

The main thing to consider is whether we should return `(AuthStatus, error)` or `(*AuthStatus, error)`. I generally like returning pointers (and nil if there is no data), but since only Docker API versions >= 1.23 will return a struct that would be non-nil in this implementation, I think a non-pointer might be better. I want to default to no panics for nil access over what would be a silent error (ie the fields just won't be filled out).

What do you think? Also what do you think of the name `AuthStatus`?

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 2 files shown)


fsouza:
@peter-edge using a non-pointer is safe-enough. We could never return ``nil, nil`` as well, but then the code would look almost the same, but with ``return &authStatus, nil`` instead of just ``return authStatus, nil``.

Regarding the name, I'm not sure I can come up with a better alternative. It's a response to the authenticate request, but calling it AuthResponse doesn't sound good.

Thanks for contributing!

PR #510, opened 2016-04-19
Fix JSON property name for IPAMOptions.Config.

vlad-alexandru-ionescu:
See for example [docs here](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.23/#2-5-networks) (`[0].IPAM.Config`).

This was causing the `NetworkInfo` function to return empty config every time.


peter-edge:
LGTM

fsouza:
@vlad-alexandru-ionescu thanks!

vlad-alexandru-ionescu:
@fsouza Thanks for quick merge!

PR #512, opened 2016-04-20
Add inactivityTimeout for all streamable api calls

cezarsa:
With this pull request, if no data is received in `inactivityTimeout` time the request will be canceled. This is specially useful for Pull And Push on a unreliable network that could start to silently drop packages between two hosts.


fsouza:
That's a nice approach, thanks for working on that and for including tests! :D

Issue #511, opened 2016-04-20
data race in events monitoring code


cezarsa:
I could only get this race trace when running the tests with Go tip. It seems like a write on closed channel panic is possible after taking a quick look at this.

```
$ go test -race
==================
WARNING: DATA RACE
Write by goroutine 80:
  runtime.closechan()
      /Users/cezarsa/code/go/src/runtime/chan.go:303 +0x0
  github.com/fsouza/go-dockerclient.(*eventMonitoringState).disableEventMonitoring()
      /Users/cezarsa/go/src/github.com/fsouza/go-dockerclient/event.go:184 +0x112
  github.com/fsouza/go-dockerclient.(*Client).RemoveEventListener()
      /Users/cezarsa/go/src/github.com/fsouza/go-dockerclient/event.go:111 +0x109
  github.com/fsouza/go-dockerclient.testEventListeners.func2()
      /Users/cezarsa/go/src/github.com/fsouza/go-dockerclient/event_test.go:235 +0x51
  github.com/fsouza/go-dockerclient.testEventListeners()
      /Users/cezarsa/go/src/github.com/fsouza/go-dockerclient/event_test.go:258 +0x187e
  github.com/fsouza/go-dockerclient.TestTLSEventListeners()
      /Users/cezarsa/go/src/github.com/fsouza/go-dockerclient/event_test.go:50 +0x8a
  testing.tRunner()
      /Users/cezarsa/code/go/src/testing/testing.go:547 +0xbf

Previous read by goroutine 17:
  runtime.chansend()
      /Users/cezarsa/code/go/src/runtime/chan.go:122 +0x0
  github.com/fsouza/go-dockerclient.(*Client).eventHijack.func1()
      /Users/cezarsa/go/src/github.com/fsouza/go-dockerclient/event.go:315 +0x489

Goroutine 80 (running) created at:
  testing.(*T).Run()
      /Users/cezarsa/code/go/src/testing/testing.go:578 +0x385
  testing.RunTests.func1()
      /Users/cezarsa/code/go/src/testing/testing.go:724 +0xa0
  testing.tRunner()
      /Users/cezarsa/code/go/src/testing/testing.go:547 +0xbf
  testing.RunTests()
      /Users/cezarsa/code/go/src/testing/testing.go:730 +0x448
  testing.(*M).Run()
      /Users/cezarsa/code/go/src/testing/testing.go:675 +0x125
  main.main()
[...]

PR #514, opened 2016-04-22
Fix send on closed channel when re-adding event listener

robbertkl:
When the last event listener is removed, the internal event channel (`eventMonitoringState.C`) is closed, but the running goroutine from eventHijack will remain running, waiting on the docker daemon for new events. As soon as a new event arrives from the docker daemon, the goroutine returns because it detects event monitoring is no longer enabled.

However, when a new event listener is added (or the old one re-added) before a new event arrives, the old goroutine will think everything is fine because monitoring is enabled. It will then send the event to the old, now closed, internal event channel, resulting in a "send on closed channel" panic.

This PR addresses this issue by checking not only if event monitoring is enabled, but also if the internal event channel is still the same one as the goroutine is using. If not, stop the goroutine.

I've included a test case which covers this specific case.

### DIFF
--- a/event.go
+++ b/event.go
@@ -310,7 +310,7 @@ func (c *Client) eventHijack(startTime int64, eventChan chan *APIEvents, errChan
 			var event APIEvents
 			if err = decoder.Decode(&event); err != nil {
 				if err == io.EOF || err == io.ErrUnexpectedEOF {
-					if c.eventMonitor.isEnabled() {
+					if c.eventMonitor.isEnabled() && c.eventMonitor.C == eventChan {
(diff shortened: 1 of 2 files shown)


fsouza:
Nice fix @robbertkl, thanks for contributing! I suppose this solves #511?

robbertkl:
To be honest, I'm not sure.. There could be other race conditions leading to closed channels being sent to. This is just a case I ran into myself. I'll comment on #511 to mention this.

fsouza:
Ok then, I'll merge this one and dig deeper in #511. Thank you for contributing!

PR #515, opened 2016-04-26
testing: lock before find on removal, otherwise index may not exist

cezarsa:
I think this problem also exist on some other handlers using the `index` return argument from `findContainer` however this problem is more evident in `removeContainer` as it's quite easy to trigger a index out of range panic with some simultaneous removals.

### DIFF
--- a/testing/server.go
+++ b/testing/server.go
@@ -718,7 +718,9 @@ func (s *DockerServer) waitContainer(w http.ResponseWriter, r *http.Request) {
 func (s *DockerServer) removeContainer(w http.ResponseWriter, r *http.Request) {
 	id := mux.Vars(r)["id"]
 	force := r.URL.Query().Get("force")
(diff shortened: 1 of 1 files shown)


testing/server.go
```diff
@@ -729,8 +731,6 @@ func (s *DockerServer) removeContainer(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.WriteHeader(http.StatusNoContent)
-	s.cMut.Lock()
-	defer s.cMut.Unlock()
 	if s.containers[index].ID == id || s.containers[index].Name == id {
```
fsouza: Since the mutex is still locked, you should not need this if anymore.

cezarsa: You're right. Fixed!

fsouza:
Thanks! :-D

PR #520, opened 2016-05-19
Add MemoryReservation to HostConfig struct

carsonoid:
Setting MemoryReservation must be done via HostConfig not Config. It's ok to leave the MemoryReservation field in the Config stuct for backward compatibility and to not break the api because the docker api will just ignore it anyway.


fsouza:
LGTM, thanks!

PR #533, opened 2016-06-13
Use DOCKER_CONFIG env var for auth. Fixes #530

### LINKED ISSUE #530: Support DOCKER_CONFIG environment variable
The docker cli supports the `DOCKER_CONFIG` environment variable to indicate where the client should look for docker configuration vs the default `~/.docker`). See [the docs](https://docs.docker.com/v1.8/reference/commandline/cli/#configuration-files)

`go-dockerclient` should support it too, and/or passing in the docker config path explicitly.

margueritepd:
As described in https://github.com/fsouza/go-dockerclient/issues/530, the official docker client supports setting the environment variable `DOCKER_CONFIG` as the location in which to look for `config.json` which contains registry authentication.

This change adds that functionality to `go-dockerclient`.

It also adds a test for the function that changed.

I don't much like what I had to do for the test. My background is dynamic programming, where stubbing file reads and environment variable lookups is easy, so the result feels quite heavy-handed to me. If you have feedback regarding how I can improve the test, do comment and I'll do my best to follow it.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -46,21 +46,45 @@ type dockerConfig struct {
 	Email string `json:"email"`
 }
 
-// NewAuthConfigurationsFromDockerCfg returns AuthConfigurations from the
-// ~/.dockercfg file.
-func NewAuthConfigurationsFromDockerCfg() (*AuthConfigurations, error) {
-	var r io.Reader
-	var err error
-	p := path.Join(os.Getenv("HOME"), ".docker", "config.json")
-	r, err = os.Open(p)
+// NewAuthConfigurationsFromFile returns AuthConfigurations from a path containing JSON
+// in the same format as the .dockercfg file.
+func NewAuthConfigurationsFromFile(path string) (*AuthConfigurations, error) {
+	r, err := os.Open(path)
 	if err != nil {
-		p := path.Join(os.Getenv("HOME"), ".dockercfg")
-		r, err = os.Open(p)
-		if err != nil {
-			return nil, err
-		}
+		return nil, err
 	}
 	return NewAuthConfigurations(r)
+}
+
+func cfgPaths(dockerConfigEnv string, homeEnv string) []string {
+	var paths []string
+	if dockerConfigEnv != "" {
+		paths = append(paths, path.Join(dockerConfigEnv, "config.json"))
+	}
+	if homeEnv != "" {
+		paths = append(paths, path.Join(homeEnv, ".docker", "config.json"))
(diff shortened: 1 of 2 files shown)


auth.go
```diff
@@ -51,16 +51,27 @@ type dockerConfig struct {
 func NewAuthConfigurationsFromDockerCfg() (*AuthConfigurations, error) {
```
robottaway: This function is conflating finding the correct path to use, with opening said path and creating the NewAuthConfiguration struct. I would split this logic for finding the path out into FindConfigPath or the like and then this function becomes simpler and more single purpose in nature. Testing might feel a bit less icky also as you can write the appropriate tests around FindConfigPath and have them sprawl a bit less as well as only require two test around the NewAuthConfigurationsFromDockerCfg (one fail on path one works on path).

auth_test.go
```diff
@@ -7,11 +7,64 @@ package docker
...
+		configPath      string
+		authKey         string
+	}{
+		{tmpDir, path.Join(tmpDir, "config.json"), "DOCKER_CONFIG_config.json_test"},
+		{"", path.Join(tmpDir, ".docker", "config.json"), "HOME_config.json_test"},
+		{"", path.Join(tmpDir, ".dockercfg"), "HOME_dockercfg_test"},
+	}
+	for _, tt := range testData {
+		os.Setenv("DOCKER_CONFIG", tt.dockerConfigEnv)
+		content := fmt.Sprintf("{\"auths\":{\"%s\": {\"auth\": \"%s\"}}}", tt.authKey, auth)
```
ranjib: may be use json package and Marshal it. can use anonymous struct

margueritepd:
@fsouza I'll wait for your feedback before following the advice of my colleagues :)

fsouza:
I apologize for the delayed response. You could have two functions: one that reads the content from a given file (maybe ``NewAuthConfigurationsFromFile(filePath string)``), and make ``NewAuthConfigurationsFromDockerCfg()`` call the first function using the path specified by ``DOCKER_CONFIG`` or the default path (``$HOME/.docker``). Ideally we would use different names, but we should not break the library API.

margueritepd:
@fsouza Thanks for the response! I've updated the PR. I'm much happier with the tests. LMK if there's anything else I should do before this is mergeable.

fsouza:
LGTM, thanks for contributing!

Travis is too slow today, I'm not going to wait for it. Tests passed locally.

PR #538, opened 2016-06-21
Set timeout function

dadgar:
This PR allows a timeout to be set on both the TCP and Unix Socket HTTP client


fsouza:
LGTM, thanks!

PR #540, opened 2016-06-23
Implement testing server docker remote api version endpoint

emosbaugh:
The following commit breaks the testing server start container endpoint:
https://github.com/fsouza/go-dockerclient/commit/67c30ae246a11510ed1c3b72145e2daf76e4d664
```go
	if c.serverAPIVersion == nil {
		c.checkAPIVersion()
	}
	if c.serverAPIVersion != nil && c.serverAPIVersion.LessThan(apiVersion124) {
		opts = doOptions{data: hostConfig, forceJSON: true}
	}
```

The request to `checkAPIVersion()` returns 404, setting `c.serverAPIVersion = nil`, and start container endpoint returns `API error (500): EOF` when trying to unmarshal nil HostConfig.

https://github.com/fsouza/go-dockerclient/blob/9d5fef7969a53796f7405827c3e68804f533b1ee/testing/server.go#L561-L565
```go
	err = json.NewDecoder(r.Body).Decode(&hostConfig)
	if err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
```

### DIFF
--- a/testing/server.go
+++ b/testing/server.go
@@ -145,6 +145,7 @@ func (s *DockerServer) buildMuxer() {
 	s.mux.Path("/volumes/{name:.*}").Methods("GET").HandlerFunc(s.handlerWrapper(s.inspectVolume))
 	s.mux.Path("/volumes/{name:.*}").Methods("DELETE").HandlerFunc(s.handlerWrapper(s.removeVolume))
 	s.mux.Path("/info").Methods("GET").HandlerFunc(s.handlerWrapper(s.infoDocker))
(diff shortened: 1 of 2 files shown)


emosbaugh:
the tests are failing but it doesnt seem related to the pr

fsouza:
Hi @emosbaugh, nice catch and thanks for contributing.

The change looks good to me, we should look into making it configurable in the future (i.e. if I set a newer version, start breaks if HostConfig is provided). And start should work with a nil HostConfig as well, but we definitely need the /version endpoint.

Thank you!

PR #550, opened 2016-07-16
the url must be coded

yiqguo:
I delete my local project，and then modify to resubmit again.And in the local test without any error.


fsouza:
The test passes for this one. Can I reopen and merge the PR?

PR #553, opened 2016-07-23
Add comment about deprecation of registry parameter

radeksimko:
I was wondering why the `Registry` parameter had absolutely no effect in my context and why it's not even documented in the Docker Remote API reference, then I went back in time, walked through some old versions and realised it was just removed at one point.

Hopefully this will save folks reading the code time if nothing else.
I guess the library could stop sending this parameter completely to the Remote API for certain versions, but I'm not sure about the intentions of this library in regards to backward compatibility.

-----

The same parameter in `PushImageOptions` is _actually_ unused from what I can see, check the oldest available Remote API Reference: https://docs.docker.com/v1.4/reference/api/docker_remote_api_v1.11/#push-an-image-on-the-registry - there is zero notion of that parameter.

### DIFF
--- a/image.go
+++ b/image.go
@@ -280,8 +280,12 @@ func (c *Client) PushImage(opts PushImageOptions, auth AuthConfiguration) error 
 // See https://goo.gl/iJkZjD for more details.
 type PullImageOptions struct {
 	Repository string `qs:"fromImage"`
-	Registry   string
 	Tag        string
+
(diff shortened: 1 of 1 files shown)


fsouza:
Hey @radeksimko, sorry for the looooong delay in the getting back to you, and thanks for contributing.

I _think_ these fields come from pre-1.0 Docker, but I'm not sure exactly when it existed and was used in Docker.

Issue #555, opened 2016-07-30
Docker 1.12 support


kyhavlov:
Are there plans to update this repo to support the new API endpoints like node/service in 1.12?

andrestc:
I think we are only missing the [plugins](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/3-6-plugins-experimental) api. Should we add experimental features to the client or wait for them to become stable?

fsouza:
@andrestc we had experimental features in the past. We can wait until someone needs it.

gpolaert:
Hello, I'm faced to an issue with the 1.12 (API 1.24) when I use `AttachToContainer`.
`2016/12/22 15:26:37 Unrecognized input header: 68`

When do you plan to support it :)?
Thanks, Guillaume

gpolaert:
Sorry not sure it's related (downgrading to 1.23 didn't fix it)

fsouza:
Hey @gpolaert, this should work with Docker 1.12 already.

Can you share a snippet of code so I can take a look?

Thanks!

ericwoodruff:
I'm looking for docker 1.13 cpu-rt-period support.

fsouza:
@ericwoodruff can you open an issue for Docker 1.13? Since Docker 1.12 is supported, I'm closing this issue.

PR #560, opened 2016-08-10
Swarm: Add APIS (Init, Join, Leave and Update)

andrestc:
This PR adds support to the swarm api: init, join, leave and update.

related to #555

### DIFF
--- a/swarm.go
+++ b/swarm.go
@@ -14,14 +14,14 @@ import (
 	"github.com/docker/engine-api/types/swarm"
 )
 
-type SwarmInitOptions struct {
+type InitSwarmOptions struct {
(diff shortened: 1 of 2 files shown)


swarm.go
```diff
@@ -0,0 +1,108 @@
...
+// license that can be found in the LICENSE file.
+
+package docker
+
+import (
+	"encoding/json"
+	"net/url"
+	"strconv"
+
+	"golang.org/x/net/context"
```
andrewsmedina: this package should be grouped with the external packages bellow

fsouza:
LGTM, thanks!

Sorry for the delayed review.

PR #569, opened 2016-08-19
make client endpoint public

upccup:


upccup:
@fsouza there is problem that i have more than one docker daemon server, but i want use the same client config. so i want to make the client endpoint to public, and i can set the endpoint before visit appoint docker server.

this is only a suggest.

fsouza:
@upccup I don't know if I like the idea, as it tends to be racy. Maybe we can solve this elsewhere? What client config you need to share?

upccup:
@fsouza  i want to share the enpoint beacause there has more than one docker daemon server i need to visit

fsouza:
@upccup but why can't you create one instance of the client per daemon?

upccup:
@fsouza  if i create create one instance of the client per daemon that mean i must cache the client with each daemon id/name  but the map of endpoint and id/name is not Stabled

fsouza:
Oh, for some reason I lost this message. Sorry about that. How are you representing your daemon?

If the daemon can be represent as a struct, the client can be embedded in this struct, and you could get the client through a method, creating instances on demand and caching if you want. Similar to what docker-cluster does: https://github.com/tsuru/docker-cluster/blob/b213b4594c6afa048aa554c1073670d05babd11f/cluster/node.go#L108.

fsouza:
I'll close this PR, as this is probably something that is not going to be merged. Thanks for contributing!

PR #571, opened 2016-08-26
node/task: add support for node and tasks apis

andrestc:
Adds support to the node and tasks apis, introduced in docker 1.12. Related to #555


fsouza:
LGTM, thanks!

PR #575, opened 2016-09-01
testing: prevent panic on nil EndpointSpec

andrestc:
This prevents a panic on `serviceCreate`, when it's called with a nil `swarm.EndpointSpec` on the test server.


andrewsmedina:
@andrestc thank you!

Issue #531, opened 2016-06-11
Beta Docker Windows support


pietv:
A question: does dockerclient support the Beta Docker for Windows which uses the endpoint

```
endpoint := "npipe:////./pipe/docker_engine"
```
instead of 
```
endpoint := "unix:///var/run/docker.sock"
```
https://github.com/docker/docker/blob/master/opts/hosts_windows.go#L6
https://github.com/docker/docker/blob/master/opts/hosts.go#L26
https://github.com/docker/docker/blob/master/opts/hosts.go#L63-L83

Thank you in advance.

fsouza:
Hi @pietv, it isn't currently supported, I could look into that, but unfortunately I don't have access to a Windows Machine with Beta Docker for Windows to test it.

Could you help testing?

pietv:
Hi @fsouza, sure! I'm happy to test it on a Windows 10 machine. Thank you!

fsouza:
Awesome, thanks! I will dig it up, prepare a pull request and notify you. I won't be able to work on that today, but I hope I can have something by Monday.

Thanks for opening the issue.

pietv:
Monday is great! Not at all! Thank you! :)

fsouza:
@pietv I'm running late :-( Unfortunately it's more complicated than I thought (I'll need a new dependency for handling named pipes). I'm still working on that.

pietv:
@fsouza thank you for the update! yes, I've noticed the `DialPipe` machinery on the Docker side which didn't look very straightforward to incorporate. No rush and thank you!

zramsay:
@fsouza have you made any progress on this front? Now that Docker for Windows is out of beta, users are asking for supported integration. Perhaps there is work @pietv or I can help with.

fsouza:
Sorry @zramsay, I haven't had any progress on that. Help is very welcome!

I wonder if we could have the code in continuous integration afterwards.

samuelkarp:
@fsouza I've started working on this and hope to have a pull request ready next week.  Is there anything specific other than just adding support that I should handle?  (For example: the endpoint defaults to `tcp://127.0.0.1:2375` on Windows but `unix:///var/run/docker.sock` elsewhere; should the Windows behavior change to default to the named pipe or stay with TCP by default?)

fsouza:
@samuelkarp thank you very much for working on that! Regarding the default endpoint, I think we can stick to Docker's behavior here. What do you think?

[...]

Issue #353, opened 2015-08-18
Run tests again real docker daemon?


marcosnils:
What do you think about testing against a real docker daemon?. We could use docker inception to have different versions of docker and version the go-client appropriately. It'd also be very helpful when upgrading the client to identify which API's are not working anymore and need to be changed. 

I can help with this if you feel it's useful.

fsouza:
That's a great idea, but I'd like to get it running in some CI infrastructure as well, but I'm not aware of any available infrastructure for these tests (I might host it somewhere private, but I don't know any public CI infrastructure where I could run integration tests against a real Docker API). Do you have anything in mind?

Thanks for raising it.

marcosnils:
@fsouza I'm using circleci (https://circleci.com/) with a very similar environment to which I've proposed here. It offers a free plan that's more than enough to accomplish what we need.

marcosnils:
@fsouza check this link also: https://circleci.com/integrations/docker

fsouza:
@marcosnils cool, I will do some testing. Do you know their policy for handling newer Docker versions?

marcosnils:
@fsouza we shouldn't care about what version they use because if we use "docker inception" strategy we can test go-dockerclient against any version that we want. As long as they they have a way of running docker images that should be enough :). 

Let me know if you need any help.

compleatang:
FWIW, we test using circle for our eris tool (which heavily relies upon this lib), but are considering moving away from it due to circle using the BTRFS driver for docker which means that we are unable to test renaming and removing containers. Even with a docker inception (DinD) strategy we have been unsuccessful in getting tests to pass which perform either a rename or remove. 

Circle _can_ import newer versions of docker via a similar strategy to this -> https://github.com/eris-ltd/eris-cli/blob/master/circle.yml#L11-L12 ; as well as a DinD strategy of course. That said, neither strategy will overcome the limitations of that BTRFS driver.

marcosnils:
@fsouza ping?

fsouza:
@marcosnils I don't have anything new on that yet :-(

But I've seen that Travis added support for Docker as well: http://docs.travis-ci.com/user/docker/.

I will work on a integration test suite, so we can point it to some Docker endpoint with internet access (to enable pulling public images from Docker Hub).

fsouza:
[...]

PR #579, opened 2016-09-15
Add support for Windows named pipes

samuelkarp:
Fixes https://github.com/fsouza/go-dockerclient/issues/531

This series of commits adds support for communicating with Docker over a Windows named pipe.

There are a few things to note:
* I've only tested this with toy programs so far.
* All the unit/integ tests in this package are passing on both Linux and Windows.
* I haven't ported all the tests in `container_unix_test.go` to Windows; I started running into some issues with `Accept()` on a `net.Listener` for a named pipe and haven't spent time troubleshooting it yet.  If this is a blocker to accepting the pull request I can spend more time here.
* I did not add TLS support for named pipes.  I don't plan to do this since I don't expect to have a use-case for this.
* A few tests that were otherwise platform-independent made some assumptions about running on Linux that don't appear to always work on Windows; I've fixed them to ensure that they run on Linux and Windows.

All the commits here are contributed under the terms of the [BSD 2-clause license](https://github.com/fsouza/go-dockerclient/blob/01804de/LICENSE).

Let me know if you need anything else!

### DIFF
--- a/client_windows_test.go
+++ b/client_windows_test.go
@@ -1,0 +1,43 @@
+// +build windows
+// Copyright 2016 go-dockerclient authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package docker
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"sync"
+
+	"github.com/Microsoft/go-winio"
+)
+
+const (
+	nativeProtocol     = namedPipeProtocol
+	nativeRealEndpoint = "npipe:////./pipe/docker_engine"
+	nativeBadEndpoint  = "npipe:////./pipe/godockerclient_test_echo"
+)
+
+var (
+	// namedPipeCount is used to provide uniqueness in the named pipes generated
+	// by newNativeServer
+	namedPipeCount int
+	// namedPipesAllocateLock protects namedPipeCount
+	namedPipeAllocateLock sync.Mutex
+)
+
+func newNativeServer(handler http.Handler) (*httptest.Server, func(), error) {
+	namedPipeAllocateLock.Lock()
+	defer namedPipeAllocateLock.Unlock()
+	pipeName := fmt.Sprintf("//./pipe/godockerclient_test_%d", namedPipeCount)
+	namedPipeCount++
+	l, err := winio.ListenPipe(pipeName, &winio.PipeConfig{MessageMode: true})
+	if err != nil {
+		return nil, nil, err
+	}
+	srv := httptest.NewUnstartedServer(handler)
+	srv.Listener = l
+	return srv, func() {}, nil
+}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,15 @@
+version: '{build}'
+platform: x64
+clone_depth: 2
+clone_folder: c:\gopath\src\github.com\fsouza\go-dockerclient
+environment:
+  GOPATH: c:\gopath
+  GOVERSION: 1.7.1
+install:
+  - set PATH=%GOPATH%\bin;c:\go\bin;%PATH%
+  - rmdir c:\go /s /q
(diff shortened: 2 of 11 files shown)


fsouza:
@samuelkarp thanks for you awesome work :-D

I'll try to get a Windows CI to make sure we run the tests. Do you know if we can run integration tests on AppVeyor (i.e. have a Docker daemon running in the box)?

Cheers!

samuelkarp:
@fsouza I missed the `integration` target in the Makefile; I haven't ported that test to Windows yet.  I haven't used AppVeyor, but the other tests don't require a running Docker daemon and should (theoretically) work.

Looking at `TestIntegrationPullCreateStartLogs`, this appears to assume a Linux daemon (it uses a Linux image and runs `cat`).  Since a Windows client might be used to run either Linux or Windows containers (Linux containers through "Docker for Windows" and Windows containers through "Docker on Windows") but the daemon would be different, should the test assume that it can use a Windows container on Windows or should it do something like query `docker info` to find the `OSType` and use that as switching logic?

fsouza:
@samuelkarp hm, I'm not entirely sure. Ideally it would test that it can communicate with named pipes, so we would need a whole new integration test. Here's what I'm going to do: I'll setup the appveyor to run ``make test`` and ensure that everything works, along with reviewing the code itself, and then I'll drop another issue for running integration tests on Windows (or simply add a note to #353). What do you think?

samuelkarp:
@fsouza That sounds like a good plan.

Testing whether it can communicate over named pipes should be easy (the other tests already do that and we can either pass the named pipe location to `NewClient` or call `NewClientFromEnv` to get the named pipe automatically), but both "Docker for Windows" and "Docker on Windows" use named pipes so the question of what the test should be is still an open one.

I can't promise that I'll have time to get the integration test done, but I'll see if I can get something quickly today.

zramsay:
yay! thanks for doing this @samuelkarp. very helpful :+1:  

@fsouza "Do you know if we can run integration tests on AppVeyor" => we're currently using docker-machine for CI on windows, but we'll definitely be investigating getting Docker for Windows up and running on AppVeyor.

client_windows_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+package docker
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"sync"
+
+	winio "github.com/Microsoft/go-winio"
```
zramsay: does the package need the name declaration? `client_windows.go` doesn't have it

samuelkarp: No, not necessary.  I think `goimports` did it for me here and I must have manually written the import in the other file.

fsouza: Nice catch @zramsay, thanks!

@samuelkarp can you remove the aliasing from this line and make this just "github.com/Microsoft/go-winio".

fsouza:
@samuelkarp changes looks mostly good to me. Sorry for the long delay in getting back to you.

I have appveyor setup in a branch with this PR merged, the test suite passes, but the race detector detected some data races: https://ci.appveyor.com/project/fsouza/go-dockerclient/build/6.

Can you take a look? It looks like calling close concurrently to any write/read is not safe :-(

You can copy the appveyor.yml file to your branch to have app veyor running your changes in the PR.

samuelkarp:
@fsouza Thank you for taking a look!  I've addressed the aliasing comment and opened https://github.com/Microsoft/go-winio/pull/31 for the race condition in the go-winio library.

samuelkarp:
(I've also rebased on top of the current tip of `master` since there were some merge conflicts that needed to be resolved.)

fsouza:
@samuelkarp awesome, thanks! I'll just wait for travis and merge this. Before pushing to master, I'll disable ``-race`` in app veyor, and keep an eye on Microsoft/go-winio#31.

Thank you very much for working on that!

PR #582, opened 2016-09-20
testing: add basic swarm support for multiple nodes

cezarsa:
This pull request add support to multiple swarm nodes in the testing client.

A new server is started after running `/swarm/init` or `/swarm/join` respecting the ListenAddr parameter or choosing a new random localhost address.

After the cluster is started all nodes joining it are exposed for listing and manipulation using the `/nodes*` routes. Information about new nodes and updates to existing nodes are propagated to all members of the cluster using an internal route in the started swarm server.

This implementation mirrors basic functionality exposed by docker swarm mode, there are probably some corner and error cases that do not completely reflect docker behavior. However, I think it's a good enough start point to enable accepting further contributions and improve from here.

### DIFF
--- a/testing/server.go
+++ b/testing/server.go
@@ -8,6 +8,7 @@ package testing
 
 import (
 	"archive/tar"
+	"bytes"
 	"crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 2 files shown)


fsouza:
lgtm, thanks!

PR #583, opened 2016-09-21
add create volume labels

upccup:


testing/server.go
```diff
@@ -28,7 +28,7 @@ import (
 
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/engine-api/types/swarm"
-	"github.com/fsouza/go-dockerclient"
+	"github.com/Dataman-Cloud/go-dockerclient"
```
fsouza: Please change this import path.

PR #587, opened 2016-09-26
Add the InspectContainerWithContext API.

aaithal:
It extends the InspectContainer API with a context. This lets users to
utilize the 'context' field in 'doOpts' to cancel '/inspect' calls to docker.

Added tests for the same.


aaithal:
Thanks for the quick merge!

PR #604, opened 2016-11-07
add param labels for build images

1851616111:
There is a very common application scenarios. Application need list image by labels, at the same time,  it also need buid image with label。

### DIFF
--- a/image.go
+++ b/image.go
@@ -459,6 +459,7 @@ type BuildImageOptions struct {
 	ContextDir          string             `qs:"-"`
 	Ulimits             []ULimit           `qs:"-"`
 	BuildArgs           []BuildArg         `qs:"-"`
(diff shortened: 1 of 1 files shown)


image.go
```diff
@@ -525,6 +526,18 @@ func (c *Client) BuildImage(opts BuildImageOptions) error {
...
+		for k, v := range opts.Labels {
+			label[k] = v
+		}
+		if b, err := json.Marshal(label); err == nil {
+			item := url.Values(map[string][]string{})
+			item.Add("labels", string(b))
+			qs = fmt.Sprintf("%s&%s", qs, item.Encode())
+		}
+	}
+
```
fsouza: The queryString function already does that, so you just need to use ``qs:"labels"`` in the Labels field.

PR #608, opened 2016-11-23
Allow TLS without client credentials

EricMountain-1A:
This patch allows TLS to work in the absence of client credentials (certificate and key), same as the Docker client.

[Docker documentation](https://docs.docker.com/engine/security/https/#client-modes) allows use of TLS without the server/daemon authenticating the client.

We've come across this situation in our use of the go-dockerclient API, so here's a proposed patch...


fsouza:
Hi @EricMountain-1A, thanks for contributing!

Can you add a test?

EricMountain-1A:
Good point, tests added @fsouza .

fsouza:
Thanks!

PR #612, opened 2016-12-28
Add ability to use registry token authentication

rossnanop:
Adds RegistryToken to AuthConfiguration struct to match docker/api/types.AuthConfig
Reference: https://godoc.org/github.com/docker/docker/api/types#AuthConfig 

This is useful when token authentication is configured on a private registry and want to connect using go-dockerclient.  Usage below:
```go
t := "jwt token"
err := docker.PushImage(docker.PushImageOptions{
  ...}, 
  docker.AuthConfiguration{RegistryToken: t}
);
```

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -26,6 +26,7 @@ type AuthConfiguration struct {
 	Username      string `json:"username,omitempty"`
 	Password      string `json:"password,omitempty"`
 	Email         string `json:"email,omitempty"`
+	RegistryToken string `json:"registrytoken,omitempty"`
 	ServerAddress string `json:"serveraddress,omitempty"`
 }
 


rossnanop:
So after looking at docker github issues, it seems registrytoken is deprecated in favor of identitytoken.  I would like to be able to use the client with token authentication, but now am not sure what the correct way to authenticate would be.  @fsouza would it make sense to switch this to use identitytoken?  Is that all that would be required to get token auth to work?

fsouza:
Hey @rossnanop, sorry for taking so long to get back to you.

I'm not entirely familiar with token authentication on Docker, I need to do some research to figure out the best approach. Do you have any suggestions?

PR #613, opened 2017-01-04
Logs(): add note on how to parse log streams

ahmetalpbalkan:


fsouza:
Hey @ahmetalpbalkan, go-dockerclient should be able to multiplex the response (if ``RawTerminal`` is ``false``).

I must admit that I haven't _actually_ used ``Logs`` (only ``Attach``), so I want to ask: did you find any issues using Logs with ``RawTerminal=true``?

ahmetalpbalkan:
@fsouza I see it now. It looks like setting RawTerminal uses stdcopy logic directly from docker/docker. So I guess we don’t need it here. You can close it if you like.

fsouza:
I can improve the documentation to mention that when RawTerminal is set to true, go-dockerclient won't parse the output and multiplex the stream, and that dlog can be used in this case. What do you think?

ahmetalpbalkan:
Sounds very good. Thanks.

PR #617, opened 2017-01-28
image: Add NetworkMode to BuildImageOptions

kklin:
Docker API v1.25 now supports setting the network mode when building
images.


fsouza:
lgtm, thanks!

PR #619, opened 2017-02-04
use http.HandlerFunc instead explict func declaration

felipeweb:


felipeweb:
I think it is failing randomly

PR #621, opened 2017-02-11
fix error with unix socket not resolving on dial

faide:
the first example of README fails with:

`panic: Get http://unix.sock/images/json?: dial tcp: lookup unix.sock on 127.0.1.1:53: no such host`

on a Ubuntu.

After investigations it appears the Dial attribute of the hashicorp cleanhttp.DefaultTransport() is deprecated. Using the new recommanded DialContext fixes the issue...

### DIFF
--- a/client_unix.go
+++ b/client_unix.go
@@ -6,6 +6,7 @@
 package docker
 
 import (
+	"context"
 	"net"
 	"net/http"
 
@@ -20,7 +21,7 @@ func (c *Client) initializeNativeClient() {
 	}
 	socketPath := c.endpointURL.Path
 	tr := cleanhttp.DefaultTransport()
-	tr.Dial = func(network, addr string) (net.Conn, error) {
+	tr.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
 		return c.Dialer.Dial(unixProtocol, socketPath)
 	}
 	c.nativeHTTPClient = &http.Client{Transport: tr}


fsouza:
@faide that's surprising. The example on README should work fine. Can you provide more information on your environment? Is Docker bound to ``/var/run/docker.sock`` and does your user has access to it?

faide:
Hello,

I created a sample project that reproduces the issue:

[https://github.com/faide/dockerpilot](url)

In this project I used govendor to pin dependencies that were used at the time. Nothing fancy here I just pulled the latest of each dependency and pinned the versions with govendor to help reproduce.

My system is:

- Ubuntu 16.04 LTS
- Go 1.7.4 (go1.7.4 linux/amd64)
- docker version : Docker version 1.13.0, build 49bf474
- my user is allowed to used the docker daemon through CLI (I am part of the docker group, the command `groups` shows it)
- the docker unix socket exists and is writable for the docker group: (srw-rw---- 1 root docker 0 févr. 12 17:04 /var/run/docker.sock)

faide:
to reproduce:

install govendor: `go get -u github.com/kardianos/govendor`

checkout master of the demo project (Without the PR applied to the go-dockerclient), build and run: you get an error:
`panic: Get http://unix.sock/images/json?: dial tcp: lookup unix.sock on 127.0.1.1:53: no such host`

checkout the pr-activated branch of the demo project (the current PR is applied to the vendored go-dockerclient) `govendor sync` to get the correct vendor tree and rebuild the demo project... it works and can enumerate the images

faide:
@fsouza this commit is the source of this issue: https://github.com/hashicorp/go-cleanhttp/commit/a45970658e51fea2c41445ff0f7e07106d007617

as you can see the latest go-cleanhttp is not using transport.Dial anymore but transport.DialContext, thus the code was changing the false Dialer and was having no effect if you were using the latest go-cleanhttp revision.

I suspect this is the case only with go-1.7+ (as go-1.6 and lower are just broken by the latest go-cleanhttp revision and cannot work anymore)

This commit is from feb 11, I must have stumbled on it just a few minutes after it was committed :)

faide:
@fsouza btw this means that all test in the matrix that are using go < 1.7 will fail with the newer version of hashicorp cleanhttp. [According to this](https://github.com/hashicorp/go-cleanhttp/commit/a45970658e51fea2c41445ff0f7e07106d007617) you can just drop support for go 1.6

I do not get why I have a failure on [1.7 with the i386 build with docker 1.11.2](https://travis-ci.org/fsouza/go-dockerclient/jobs/200756032). This appears only on i386 on this specific version of docker... strange.

kcmerrill:
@faide thanks for that PR! I was wondering what was going on ...

@fsouza not sure if this helps in the troubleshooting, but master works when the unix socket is local, but it didn't work when I was using the unix socket inside a container. I had everything vendored too. Hopefully that's not a red herring but it's a problem I observed.

ewoutp:
This fix works fine with docker on Mac.

fsouza:
I see what's going on, this is also related to #623.

I'm _probably_ going to drop Go 1.6 and merge this PR. @faide how urgent is this for you at the moment?

Thanks for contributing!

fsouza:
Quick note: I decided to merge this after reading the reports on #622.

PR #634, opened 2017-02-28
add support for cgroup parent on image build

bparees:
the api does not appear to be documented but it's implemented here:
https://github.com/docker/docker/blob/master/api/server/router/build/build_routes.go#L53

and the docker-build cli tool leverages it.

### DIFF
--- a/image.go
+++ b/image.go
@@ -462,6 +462,7 @@ type BuildImageOptions struct {
 	BuildArgs           []BuildArg         `qs:"-"`
 	NetworkMode         string             `qs:"networkmode"`
 	InactivityTimeout   time.Duration      `qs:"-"`
+	CgroupParent        string             `qs:"cgroupparent"`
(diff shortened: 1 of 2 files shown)


bparees:
@smarterclayton @derekwaynecarr fyi.

image.go
```diff
@@ -463,6 +463,7 @@ type BuildImageOptions struct {
 	NetworkMode         string             `qs:"networkmode"`
 	InactivityTimeout   time.Duration      `qs:"-"`
 	Context             context.Context
+	CgroupParent        string `qs:"cgroupparent"`
```
fsouza: Can you move this up so the Context is the last field in the struct?

Thanks!

fsouza:
Hey @bparees, thanks for contributing! Just a minor change,

bparees:
@fsouza sure thing, done.

PR #636, opened 2017-03-04
Private registry auth support for creating swarm mode services

xesina:
In order to create swarm services from private registries, we need to provide AuthConfiguration.


service.go
```diff
@@ -39,9 +39,14 @@ type CreateServiceOptions struct {
 // or an error in case of failure.
 //
 // See https://goo.gl/KrVjHz for more details.
-func (c *Client) CreateService(opts CreateServiceOptions) (*swarm.Service, error) {
+func (c *Client) CreateService(opts CreateServiceOptions, auth AuthConfiguration) (*swarm.Service, error) {
```
fsouza: Can you add AuthConfigure to the CreateServiceOptions struct and set it to `qs:"-"`? This way we don't break the API.

xesina: @fsouza Done. Please let me know if there is any other issues.

PR #653, opened 2017-05-21
Include on Docker Inspect: Container network aliases

jwayong:
When doing docker inspect, there is a Aliases information on NetworkSettings.Networks, this is a valueable information that can be used. This should be supported for API 1.23 above.

Example output JSON from Docker inspect:

```
"someusernetwork": {
     "IPAMConfig": null,
     "Links": null,
     "Aliases": [
         "web01",
         "dfcfa712b3a9"
     ],
     "NetworkID": "751d96e988b498284bd1ddf940295787691afdcc73dc8c63bbbca5dda2a31752",
     "EndpointID": "4610d818cf63b73237d411e43a8d4a79ca9df8f1dae9a271daf0623c4f3c78e7",
     "Gateway": "172.18.0.1",
     "IPAddress": "172.18.0.2",
     "IPPrefixLen": 16,
     "IPv6Gateway": "",
     "GlobalIPv6Address": "",
     "GlobalIPv6PrefixLen": 0,
     "MacAddress": "02:42:ac:12:00:02"
 }
```

### DIFF
--- a/container_test.go
+++ b/container_test.go
@@ -629,8 +629,8 @@ func TestInspectContainerNetwork(t *testing.T) {
 	id := "81e1bbe20b55"
 	expIP := "10.0.0.3"
 	expNetworkID := "7ea29fc1412292a2d7bba362f9253545fecdfa8ce9a6e37dd10ba8bee7129812"
-	expAlias0 = "testalias"
-	expAlias1 = "81e1bbe20b55"
+	expAlias0 := "testalias"
+	expAlias1 := "81e1bbe20b55"
 
 	container, err := client.InspectContainer(id)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


container_test.go
```diff
@@ -652,6 +658,17 @@ func TestInspectContainerNetwork(t *testing.T) {
...
 			}
 		}
+
+		var aliases []string
+		for _, net := range networks.MapKeys() {
+			if net.Interface().(string) == container.HostConfig.NetworkMode {
+				aliases = networks.MapIndex(net).FieldByName("Aliases").Interface().([]string)
+			}
+		}
+		if aliases[0] != expAlias0 || aliases[1] != expAlias1 {
```
fsouza: Can you change this check to use ``reflect.DeepEqual``? Something like:

```go
expectedAliases := []string{"testalias", "81e1bbe20b55"}
// ...
if !reflect.DeepEqual(aliases, expectedAliases) {
  t.Errorf("...", ...)
}
```

With this you can remove the variables expAlias0 and expAlias1.

jwayong: Thanks for the feedback, it make sense. I've done the change.

jwayong:
Thanks :)

PR #655, opened 2017-06-26
Env create exec option

dmtsui:
Currently CreateExecOption doesn't support adding environmental variables that is supported in api ver 1.25. This pull request is to add support for the Env configuration for Exec. Error handling has been added to notify <1.25 api users that Env is not supported for CreateExecOption.


dmtsui:
thanks for the quick review!

PR #660, opened 2017-07-24
add service logs

mackristof:
add func LogsService() & type LogsServiceOptions

### DIFF
--- a/service.go
+++ b/service.go
@@ -190,7 +190,7 @@ type LogsServiceOptions struct {
 	Details     bool
 }
 
-// LogsService gets stdout and stderr logs from the specified service.
+// GetServiceLogs gets stdout and stderr logs from the specified service.
 //
 // When LogsServiceOptions.RawTerminal is set to false, go-dockerclient will multiplex
 // the streams and send the containers stdout to LogsServiceOptions.OutputStream, and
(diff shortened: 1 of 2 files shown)


service.go
```diff
@@ -167,3 +169,48 @@ func (c *Client) ListServices(opts ListServicesOptions) ([]swarm.Service, error)
...
+
+// LogsService gets stdout and stderr logs from the specified service.
+//
+// When LogsServiceOptions.RawTerminal is set to false, go-dockerclient will multiplex
+// the streams and send the containers stdout to LogsServiceOptions.OutputStream, and
+// stderr to LogsServiceOptions.ErrorStream.
+//
+// When LogsServiceOptions.RawTerminal is true, callers will get the raw stream on
+// LogsServiceOptions.OutputStream.
+func (c *Client) LogsService(opts LogsServiceOptions) error {
```
fsouza: Should we call it something like ``GetServiceLogs`` instead? Just to keep the method starting with a verb (calling "Logs" just "Logs" is a mistake we can't fix now :/).

mackristof:
tests added on GetServiceLogs()

service.go
mackristof: done

PR #665, opened 2017-08-02
Upgrade x/net/context to use the builtin context package

### LINKED ISSUE #658: Replace "golang.org/x/net/context" with standard library "context"
Go 1.7 has moved the `golang.org/x/net/context` to standard library, and we want to get rid of the `golang.org/x/net/context` by using the standard `context`, but go-docker-client has a dependency on `golang.org/x/net/context` which is blocking us from doing so. I was wondering can go-docker-client upgrade to use `context` instead of the one under `x/net`, is there any concerns to do so?

richardpen:
Fix #658

### DIFF
--- a/client.go
+++ b/client.go
@@ -10,6 +10,7 @@ package docker
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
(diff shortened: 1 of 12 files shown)


fsouza:
Hey @richardpen, thanks for contributing.

The problem here is that this is a hard issue to solve, as it would be a breaking change. Hopefully, with type aliases it won't make any different on Go 1.9+, so I'm going to close this PR and add a comment to #658 to revisit this once Go 1.10 is released.

PR #671, opened 2017-09-29
export NativeHTTPClient on docker.Client struct

LarsFronius:
Use-case
===

I want to trace calls made to the docker unix socket or http/tcp endpoint.
Thus I plan on overwriting the [Roundtripper](https://golang.org/pkg/net/http/#RoundTripper) on the Transport of the http.Client of the docker client. While I can do this for the exported "HTTPClient" already, I can not do this on the unexported "nativeHTTPClient" used for Unix socket connections.

I don't see another way of passing a different HTTP Transport on the nativeHTTPClient in right now, other than having the one initialized exported and accessible from the outside. I am open to any ideas here.

Tests pass
---
```
~/go/src/github.com/fsouza/go-dockerclient (master) system  ᕕ( ᐛ )ᕗ   make test
go get -d -t ./...
[ -z "$(golint . | grep -v 'type name will be used as docker.DockerInfo' | grep -v 'context.Context should be the first' | tee /dev/stderr)" ]
go vet ./...
[ -z "$(gofmt -s -d . | tee /dev/stderr)" ]
go test  ./...
ok  	github.com/fsouza/go-dockerclient	2.749s
ok  	github.com/fsouza/go-dockerclient/testing	1.346s
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -151,7 +151,7 @@ type Client struct {
 	requestedAPIVersion APIVersion
 	serverAPIVersion    APIVersion
 	expectedAPIVersion  APIVersion
-	nativeHTTPClient    *http.Client
+	NativeHTTPClient    *http.Client
 }
 
 // Dialer is an interface that allows network connections to be dialed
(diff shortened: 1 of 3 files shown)


fsouza:
I wonder if we can get rid of nativeHTTPClient and have just HTTPClient. Having both can be confusing. Can you take a look at #672 and let me know what you think?

Thanks for contributing!

LarsFronius:
Agreed. I guess instead of configuring a secondary Client for socket operations, we can do that directly on the HTTPClient and thus don't need to route on the request itself to two different HTTPClients, depending on their endpoint.

fsouza:
Cool. I'll go ahead and merge #672 and close this PR. Thank you very much for contributing!

PR #674, opened 2017-10-04
add support for distribution endpoint

jeinwag:
Adds support for DistributionInspect endpoint:
https://docs.docker.com/engine/api/v1.31/#operation/DistributionInspect


johanbrandhorst:
Merging this meant `master` is no longer compatible with the latest release of docker: [`17.05.x`](https://github.com/moby/moby/blob/17.05.x/api/types/registry/registry.go). Is it the intention of this package to remain compatible only with the master branch of Docker?

PR #684, opened 2017-10-25
Add golang/dep support

### LINKED ISSUE #683: Add golang/dep support
Hi, can we add support for [golang/dep](https://github.com/golang/dep)?

Recently, we got an issue https://github.com/golang/dep/issues/1274, where a project that depended on go-dockerclient tried using dep to manage the dependencies and dep installed docker version `1.13.1`, which doesn't work with the latest go-dockerclient. This happened because dep found that version as the most recent, valid semver version of docker. The new releases of docker are of the form `v17.05.0-ce` which isn't a valid semver but a tag.

I can create a PR if we agree to add dep support 🙂

Thanks!

darkowlzz:
The project constraints are all set to master branch of the dependencies, which is the same as `go get`-ting them. I would recommend doing semver releases in the future with properly defined constraints.

fixes #683

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,0 +1,24 @@
+
+[[constraint]]
+  name = "github.com/Microsoft/go-winio"
+  branch = "master"
+
+[[constraint]]
+  name = "github.com/docker/docker"
+  branch = "master"
+
+[[constraint]]
+  name = "github.com/docker/go-units"
+  branch = "master"
+
+[[constraint]]
+  name = "github.com/google/go-cmp"
+  branch = "master"
+
+[[constraint]]
+  name = "github.com/gorilla/mux"
+  branch = "master"
+
+[[constraint]]
+  name = "golang.org/x/net"
+  branch = "master"
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ lint:
 	[ -z "$$(golint . | grep -v 'type name will be used as docker.DockerInfo' | grep -v 'context.Context should be the first' | tee /dev/stderr)" ]
 
 vet:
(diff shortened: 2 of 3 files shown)


darkowlzz:
I've tried replacing `go get`s with `dep` to download the dependencies in CI. Since `go get` downloaded the packages in $GOPATH, and dep installs them under `vendor/` directory, I had to change the test command and ignore the tests from running in vendor/. This is the default behavior in go1.9 but just for go1.8, I've added the ignore command. Shouldn't hurt. :)

@fsouza can you help with the test failure? Seems to pass on my machine.
```
--- PASS: TestClientStreamJSONDecodeWithTerminal (0.02s)
```

fsouza:
@darkowlzz the failing test looks like a bug in the tty lib, it's supposed to be fixed here: https://github.com/Nvveen/Gotty/pull/1

Docker worked around the bug by vendoring the fork: https://github.com/moby/moby/blob/17bb1d3663f6586e83b453670526e3186bb56dd3/vendor.conf#L141

I'll merge the PR  and update the constraint configurations to use the fork.

Thank you very much for working on that! I'm definitely going to start tagging releases now that we can safely specify dependency versions :)

darkowlzz:
oh! I was still trying to figure out what went wrong 😄
You can alter the source of a project by defining `source` attribute in the constraint. Refer https://github.com/golang/dep/blob/master/Gopkg.toml#L4 for an example and https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#constraint for detailed docs.

Thanks again!

fsouza:
@darkowlzz thank _you_! :)

PR #699, opened 2017-12-07
Fix image tag and inspect using image id instead of name in testing server

cezarsa:
Previously only previously tagged images could be retagged and inspected, with this PR it's now possible to tag and inspect using only the image ID.


andrestc:
LGTM.

PR #703, opened 2018-01-16
Parse the error response as JSON if possible

knight42:
I fetched the logs of a removed container, and the API returns an error like:
```go
*docker.Error {
	Status: 404,
	Message: "{\"message\":\"No such container: syncing-archlinux\"}\n"
}
```
so I think it is reasonable to parse the error if possible.

### DIFF
--- a/client.go
+++ b/client.go
@@ -946,12 +946,20 @@ type Error struct {
 }
 
 func newError(resp *http.Response) *Error {
+	type ErrMsg struct {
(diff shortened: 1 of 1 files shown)


knight42:
@fsouza just a friendly ping to keep this on your radar.

knight42:
@fsouza ooops... I have forgotten to add a test. Should I file another PR?

PR #704, opened 2018-01-19
feat: forward via optional jump hosts.

mavogel:
## Summary
This is a feature to forward the `Docker Daemon` from a remote address to a local address on your local machine and access it there.

It's backward compatible with the existing interface and adds only optional parameters.

### Notes
- It has only been tested with non-password-protected ssh-keys.
- ATM only one jump host is supported
- I tested the implementation but if desired I can also provider you a terraform `main.tf` file where you can created a testing infra on `AWS`
- Errors on the bootstrap are returned immediately
- Errors on the forward, explicitly on copying the content of the connections on both direction need to be printed out as shown below in the `handleForwardErrors` function.

## Usage
```go
package main

import (
	"log"
	"time"

	dc "github.com/fsouza/go-dockerclient"
)

func main() {
	forwardConfig := &dc.ForwardConfig{
		JumpHostConfigs: []*dc.ForwardSSHConfig{
			&dc.ForwardSSHConfig{
				Address:        "10.0.0.1:22",
				User:           "jumpuser1",
				PrivateKeyFile: "/Users/abc/.ssh/id_rsa_jump_host1",
				Password:       "",
			},
		},
		EndHostConfig: &dc.ForwardSSHConfig{
			Address:        "20.0.0.1:22",
			User:           "endhostuser",
			PrivateKeyFile: "/Users/abc/.ssh/id_rsa_end_host",
			Password:       "",
		},
		LocalAddress:  "localhost:2376",
		RemoteAddress: "localhost:2376",
	}

	endpoint := "tcp://127.0.0.1:2376"
	cert := "/Users/abc/.docker/cert.pem"
	key := "/Users/abc/.docker/key.pem"
	ca := "/Users/abc/.docker/ca.pem"
	client, err := dc.NewTLSClientWithForward(endpoint, cert, key, ca, forwardConfig) 	
        handleForwardErrors(client)
	defer client.CloseForward()
	if err != nil {
		log.Printf("err: %s", err)
		return
	}

	info, err := client.Info()
	if err != nil {
		log.Printf("err: %s", err)
		return
	}
	log.Printf("Info: %v", info)
}

func handleForwardErrors(client *dc.Client) {
	if client != nil && client.ForwardErrors != nil {
		go func() {
			for {
				select {
				case forwardError := <-client.ForwardErrors:
					log.Printf("forward err: %s", forwardError)
				case <-time.After(3 * time.Second):
					log.Printf("NO forward err...")
				}

			}
		}()
	}
}

```

Which is basically the same as running the following command in your shell
```sh
# setup the forward
$ ssh -f -L 2376:localhost:2376 -o ExitOnForwardFailure=yes -o ProxyCommand="ssh -l jumpuser1 -i ~/.ssh/id_rsa_jump_host1 10.0.0.1 -W %h:%p" -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ~/.ssh/id_rsa_end_host endhostuser@$20.0.0.1 sleep 10
# access the forwarded docker host
$ docker --tlsverify -H 127.0.0.1:2376 info
```

## Benefits
- The ssh tunnel is fully established within the docker client and not with a shell command
- Because this libary is used in the `terraform-provider-docker` it could then be used there in the following way if the docker hosts are behind a `Bastion host` which is the jump host here:
```hcl
provider "docker" {
  version = "~> 0.1.2" 

  host      = "tcp://127.0.0.1:2376/"
  cert_path = "${pathexpand("~/.docker")}"

  forward_config {
    jump_host_configs = [
      {
        address = "10.0.0.1:22"
        user = "jumpuser"
        private_key_file = "/Users/abc/.ssh/id_rsa_jump_host1"
      },
    ]
    end_host_config {
        address = "20.0.0.1:22"
        user = "endhostuser"
        private_key_file = "/Users/abc/.ssh/id_rsa_end_host"
    }
    local_address = "localhost:2376"
    remote_address = "localhost:2376"
  }
}
```


mavogel:
@fsouza I don't think the only failing [testarch](https://travis-ci.org/fsouza/go-dockerclient/jobs/330860309) has something to do with the implementation.

fsouza:
Hey @mavogel, thanks for proposing this and working on it. I feel like this doesn't belong to go-dockerclient, but instead to a dedicated library (that could use go-dockerclient). We can make a change to go-dockerclient so expose the needed information to create such lib, but I don't think go-dockerclient should deal with establishing the tunnel.

mavogel:
Hi @fsouza it feels like your decision is final, right? I mean you're the owner. So I'll extract the code to a separate library and make a PR if there will be things that need to be exposed.

mavogel:
I make a separate simple lib [sshforward](https://github.com/mavogel/sshforward) and it's completely independent of your lib. We go with this approach. Thanks for pointing out @fsouza 😃

fsouza:
@mavogel awesome! I'm glad you were able to get it working in a separate lib! :D

PR #712, opened 2018-01-31
adding in online_cpus as of docker v17.05 stats update

sebito91:
The `CPUStats` section now includes an `online_cpus` component which is roughly (usually?) the `len(cpu_usage.percpu_usage)` list within the same struct. I'm not 100% certain this appeared as of v17.05, but it's definitely now there and part of the `cpuPercent` calculation in the docker stats api.

JSON output example -- https://pastebin.com/uhmgxHiF
moby code -- https://github.com/moby/moby/blob/17.05.x/cli/command/container/stats_helpers.go#L174

### DIFF
--- a/container.go
+++ b/container.go
@@ -15,7 +15,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/docker/go-units"
+	units "github.com/docker/go-units"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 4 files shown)


sebito91:
No problem, will do!

sebito91:
🎉

Issue #644, opened 2017-03-27
Docker plugins api support question


prateekgogia:
I checked the code and couldn't find support for plugin apis present in Docker, I am specifically looking for this- 
https://docs.docker.com/engine/api/v1.25/#tag/Plugin

Can someone please let me know if the support is added to this package ?

fsouza:
@prateekgogia hey, thanks for opening this issue and sorry for the delayed response. I'm not aware of anyone working on the plugin integration, but I'd be happy to review changes related to that. I can also work on that too, but not this week :(

fsouza:
Done via #713.

PR #713, opened 2018-02-08
[Add] plugins api #644

airking05:
Hi,
I have wrote plugins api functions for Docker API 1.25.
I would like you to review.

https://docs.docker.com/engine/api/v1.25/#tag/Plugin

### DIFF
--- a/plugin.go
+++ b/plugin.go
@@ -20,7 +20,7 @@ type PluginPrivilege struct {
 	Value       []string `json:"Value,omitempty" yaml:"Value,omitempty" toml:"Value,omitempty"`
 }
 
-// InstallPluginOptions This is a TBD Comments.
+// InstallPluginOptions .This is a TBD Comments.
 //
 // See https://goo.gl/kaOHGw for more details.
 type InstallPluginOptions struct {
@@ -51,7 +51,7 @@ func (c *Client) InstallPlugins(opts InstallPluginOptions, auth AuthConfiguratio
 	return nil
 }
 
-// PluginSetting This is a TBD Comments.
+// PluginSetting .This is a TBD Comments.
 //
 // See https://goo.gl/kaOHGw for more details.
 type PluginSetting struct {
@@ -60,7 +60,7 @@ type PluginSetting struct {
 	Devices []string `json:"Devices,omitempty" yaml:"Devices,omitempty" toml:"Devices,omitempty"`
 }
 
-// PluginInterface This is a TBD Comments.
+// PluginInterface .This is a TBD Comments.
 //
 // See https://goo.gl/kaOHGw for more details.
 type PluginInterface struct {
@@ -68,14 +68,14 @@ type PluginInterface struct {
 	Socket string   `json:"Socket,omitempty" yaml:"Socket,omitempty" toml:"Socket,omitempty"`
 }
 
-// PluginNetwork This is a TBD Comments.
+// PluginNetwork .This is a TBD Comments.
 //
 // See https://goo.gl/kaOHGw for more details.
 type PluginNetwork struct {
 	Type string `json:"Type,omitempty" yaml:"Type,omitempty" toml:"Type,omitempty"`
 }
 
-// PluginLinux This is a TBD Comments.
+// PluginLinux .This is a TBD Comments.
 //
 // See https://goo.gl/kaOHGw for more details.
 type PluginLinux struct {
@@ -84,7 +84,7 @@ type PluginLinux struct {
 	Devices         []PluginLinuxDevices `json:"Devices,omitempty" yaml:"Devices,omitempty" toml:"Devices,omitempty"`
 }
 
-// PluginLinuxDevices This is a TBD Comments.
+// PluginLinuxDevices .This is a TBD Comments.
 //
 // See https://goo.gl/kaOHGw for more details.
 type PluginLinuxDevices struct {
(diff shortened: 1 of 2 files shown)


plugin.go
```diff
@@ -0,0 +1,391 @@
...
+//
+// See https://goo.gl/kaOHGw for more details.
+type InstallPluginOptions struct {
+	Remote  string
+	Name    string
+	Plugins []PluginPrivilege
+	Context context.Context
+}
+
+// InstallPlugins returns a slice of containers matching the given criteria.
```
fsouza: Can you update this method description?

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+		context: opts.Context,
+	})
+	defer resp.Body.Close()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// PluginSetting .This is a TBD Comments.
```
fsouza: Can you add a brief description to this and other types/methods?

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+//
+// See https://goo.gl/kaOHGw for more details.
+type PluginSetting struct {
```
fsouza: Can you rename this type to ``PluginSettings``?

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+// InstallPlugins returns a slice of containers matching the given criteria.
+//
+// See https://goo.gl/kaOHGw for more details.
+func (c *Client) InstallPlugins(opts InstallPluginOptions, auth AuthConfiguration) error {
+	params := make(url.Values)
+	params.Set("remote", opts.Remote)
+	if opts.Name != "" {
+		params.Set("name", opts.Name)
+	}
+	path := "/plugins/pull?" + queryString(params)
```
fsouza: Since you're using ``queryString()``, you can add struct tags to fields on ``InstallPluginOptions`` and just do:

```go
path := "/plugins/pull?" + queryString(opts)
```

``InstallPluginOptions`` would look something like:

```go
type InstallPluginOptions struct {
	Remote  string
	Name    string
	Plugins []PluginPrivilege `qs:"-"`
	Context context.Context  
}
```

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+	Remote  string
+	Name    string
+	Plugins []PluginPrivilege
+	Context context.Context
+}
+
+// InstallPlugins returns a slice of containers matching the given criteria.
+//
+// See https://goo.gl/kaOHGw for more details.
+func (c *Client) InstallPlugins(opts InstallPluginOptions, auth AuthConfiguration) error {
```
fsouza: Can you move ``AuthConfiguration`` into ``InstallPluginOptions``?

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+		return nil, err
+	}
+	return pluginDetails, nil
+}
+
+// GetPluginPrivileges .This is a TBD Comments.
+//
+// See https://goo.gl/kaOHGw for more details.
+func (c *Client) GetPluginPrivileges(name string) ([]PluginPrivilege, error) {
+	resp, err := c.do("GET", "/plugins/privileges?"+name, doOptions{})
```
fsouza: This should be ``"/plugins/privileges?remote="+name``.

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+		return nil, err
+	}
+	return &pluginDetail, nil
+}
+
+// RemovePluginOptions .This is a TBD Comments.
+//
+// See https://goo.gl/kaOHGw for more details.
+type RemovePluginOptions struct {
+	// The ID of the container.
```
fsouza: Please fix the field description to be `` The name of the plugin.`` (we can probably get rid of this too, as it's kinda obvious).

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&pluginDetails); err != nil {
+		return nil, err
+	}
+	return pluginDetails, nil
+}
+
+// GetPluginPrivileges .This is a TBD Comments.
+//
+// See https://goo.gl/kaOHGw for more details.
+func (c *Client) GetPluginPrivileges(name string) ([]PluginPrivilege, error) {
```
fsouza: Can you  make sure that this and all other methods get a ``Context``? I tagged them with "Missing ``Context``." to make it easier for you to track them as you work on the changes. The idea is to always support canceling in flight requests if the context gets canceled.

Please use a struct (``GetPluginPrivilegesOptions``), as it makes it easier to add new options as Docker evolves their API.

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+	Tag      string        `json:"Tag,omitempty" yaml:"Tag,omitempty" toml:"Tag,omitempty"`
+	Active   bool          `json:"Active,omitempty" yaml:"Active,omitempty" toml:"Active,omitempty"`
+	Settings PluginSetting `json:"Settings,omitempty" yaml:"Settings,omitempty" toml:"Settings,omitempty"`
+	Config   PluginConfig  `json:"Config,omitempty" yaml:"Config,omitempty" toml:"Config,omitempty"`
+}
+
+// ListPlugins .This is a TBD Comments.
+//
+// See https://goo.gl/kaOHGw for more details.
+func (c *Client) ListPlugins() ([]PluginDetail, error) {
```
fsouza: Missing ``Context``.

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&pluginPrivileges); err != nil {
+		return nil, err
+	}
+	return pluginPrivileges, nil
+}
+
+// InspectPlugins .This is a TBD Comments.
+//
+// See https://goo.gl/kaOHGw for more details.
+func (c *Client) InspectPlugins(name string) (*PluginDetail, error) {
```
fsouza: Missing ``Context``.

plugin.go
```diff
@@ -0,0 +1,391 @@
...
+	Path string `qs:"-"`
+
+	Context context.Context
+}
+
+// CreatePlugin .This is a TBD Comments.
+//
+// See https://goo.gl/kaOHGw for more details.
+func (c *Client) CreatePlugin(opts CreatePluginOptions) (string, error) {
+	path := "/plugins/create?" + queryString(opts.Name)
```
fsouza: Should be ``queryString(opts)``.

airking05:
Thank you for your helping :) 
I have fixed what you pointed out and this PR is getting better i think.

PR #727, opened 2018-04-23
misc: add VersionWithContext method

aaithal:
In rare cases, Docker's "/version" API might take a long time to
respond. Wrapping this with a context gives clients more control
over handling this scenario.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 sudo: required
 go:
-  - 1.9.x
   - 1.10.x
   - tip
 os:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 sudo: required
 go:
+  - 1.9.x
   - 1.10.x
-  - tip
 os:
   - linux
   - osx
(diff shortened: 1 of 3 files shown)


aaithal:
Hi @fsouza , build seems to be broken because golint is apparently *gone*. See https://github.com/golang/lint/issues/397 for details. I'm not sure if you want to try to work around this as per https://github.com/golang/lint/issues/397#issuecomment-383752949 in this PR or in a separate PR. Please let me know.

Thanks,
Anirudh

aaithal:
FWIW, `make lint` was passing on my desktop yesterday:

```
$ make lint vet fmtcheck
[ -z "$(golint . | grep -v 'type name will be used as docker.DockerInfo' | grep -v 'context.Context should be the first' | tee /dev/stderr)" ]
go vet ./...
[ -z "$(gofmt -s -d *.go ./testing | tee /dev/stderr)" ]
```

fsouza:
@aaithal thanks for contributing, unfortunately we can't drop Go 1.9 as it's still officially supported. Can you remove that change from the .travis.yml? I can then figure out the best strategy for the ``docker/docker/pkg/archive`` change.

aaithal:
@fsouza thanks for getting back on this. I've changed it accordingly. Can you please take another look?

Thanks,
Anirudh

PR #730, opened 2018-04-30
Add Env and WorkingDir support, fix go vet for swarm tests

fayep:
In 2017.12 support was added for setting environment and working directory for execs.  This is super useful.  I've started to add support and it is working, but it doesn't test for if the version is new enough. I also saw that there were some problems with testing/swarm_test.go (had changed to accepting a pointer to the structure, but the tests weren't updated) so I fixed those too.  It might be nice to change the touch to not specify /tmp (because that's our workdir in the test now)


fayep:
Sorry, it looks as though my version was not current, I'm rebasing and it looks like it's only working dir that's been added

PR #732, opened 2018-05-01
add WorkingDir option for exec

fayep:
The ability to set working directory has been available since 17.12 (API 1.35) and it's really useful.  I've followed the example for Env when adding WorkingDir.  make test passes.  Please let me know if there are any changes you'd like me to make.


fsouza:
@fayep thank you very much for contributing!

PR #744, opened 2018-08-10
Support IdentityToken in AuthConfiguration

GeorgeMac:
This change adds support for the sending and parsing of an
IdentityToken in an AuthConfiguration. IdentityToken's are retrieved via the /auth call (see
AuthCheck) and can then be used in place of the password. After a
successful `docker login` a users `.dockercfg` can contain an identity
token in place of the original username and password in the auth field.

See: https://docs.docker.com/engine/api/v1.25/#section/Authentication

Also see official docker types https://godoc.org/github.com/docker/docker/api/types#AuthConfig

### DIFF
--- a/auth_test.go
+++ b/auth_test.go
@@ -148,6 +148,7 @@ func TestAuthAndOtherFields(t *testing.T) {
 		t.Errorf(`AuthConfigurations.Configs["docker.io"].ServerAddress: wrong result. Want %q. Got %q`, want, got)
 	}
 }
+
 func TestAuthConfig(t *testing.T) {
 	t.Parallel()
 	auth := base64.StdEncoding.EncodeToString([]byte("user:pass"))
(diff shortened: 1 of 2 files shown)


auth_test.go
```diff
@@ -174,6 +175,26 @@ func TestAuthConfig(t *testing.T) {
 	}
 }
 
+func TestAuthConfigIdentityToken(t *testing.T) {
+	t.Parallel()
+	auth := base64.StdEncoding.EncodeToString([]byte("someuser"))
+	read := strings.NewReader(fmt.Sprintf(`{"auths":{"docker.io":{"auth":"%s","identitytoken":"sometoken"}}}`, auth))
+	ac, err := NewAuthConfigurations(read)
+	if err != nil {
+		t.Error(err)
```
fsouza: Can you replace this with ``t.Fatal(err)`` so the test execution aborts after failing?

GeorgeMac:
🙇 my pleasure. I will jump on that now.

GeorgeMac:
🎉 amazing, thank you for the speedy response!

PR #745, opened 2018-08-18
Fix idtools types to follow the changes in docker library

itchyny:
This pull request fixes the following compile error, related to the changes in the docker library https://github.com/moby/moby/commit/b3e9f7b13b0f0c414fa6253e1f17a86b2cff68b5. The library added another field `SID` to `IDPair` and now it's not a pair so renamed to `Identity`. I confirmed that `make` passes.
```
# github.com/fsouza/go-dockerclient/internal/archive
internal/archive/archive.go:74:20: undefined: idtools.IDPair
```

### DIFF
--- a/internal/archive/archive.go
+++ b/internal/archive/archive.go
@@ -71,7 +71,7 @@ type TarOptions struct {
 	NoLchown         bool
 	UIDMaps          []idtools.IDMap
 	GIDMaps          []idtools.IDMap
-	ChownOpts        *idtools.IDPair
+	ChownOpts        *idtools.Identity
 	IncludeSourceDir bool
 	// WhiteoutFormat is the expected on disk format for whiteout files.
 	// This format will be converted to the standard format on pack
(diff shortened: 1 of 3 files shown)


internal/archive/archive.go
```diff
@@ -364,12 +364,12 @@ func (ta *tarAppender) addTarFile(path, name string) error {
 	//by the kernel and already have proper ownership relative to the host
 	if !isOverlayWhiteout &&
 		!strings.HasPrefix(filepath.Base(hdr.Name), WhiteoutPrefix) &&
-		!ta.IDMappings.Empty() {
-		fileIDPair, err := getFileUIDGID(fi.Sys())
+		!ta.IdentityMapping.Empty() {
+		fileIdentity, err := getFileUIDGID(fi.Sys())
```
itchyny: Should we rename this to getFileIdentity?

fsouza: Sounds good. Let's do it.

PR #749, opened 2018-09-18
added diskUsage endpoint

ericfode:


ericfode:
I did not change anything regarding the failing tests
```
--- FAIL: TestStatsTimeoutUnixSocket (0.01s)
	container_unix_test.go:60: socketPath=/tmp/socket270701668/docker_test.sock
	container_unix_test.go:99: Failed to receive timeout error, got <nil>
```
So i'm not particularly clear as to how to fix it, any suggestions would be appricatiated.

system.go
```diff
@@ -0,0 +1,90 @@
...
+	//
+	// Required: true
+	Size int64 `json:"Size"`
+}
+
+// ImageSummary represents data about what images are
+// currently known to docker
+// More Info Here https://dockr.ly/2PNzQyO
+type ImageSummary struct {
+	Containers  int64             "json:\"Containers\""
```
fsouza: Can you use `json:"Containers"` here? This way, you don't need to escape the inner quotes.

system_test.go
```diff
@@ -0,0 +1,108 @@
...
+      "Mountpoint": "",
+      "Name": "1284e17abce1d43818d7136849095c6a449a8dcfbaa859c2ff7c40abc75653eb",
+      "Options": {},
+      "Scope": "local",
+      "UsageData": {
+        "RefCount": 0,
+        "Size": 0
+      }
+    }
+  ],
```
fsouza: Can you include BuildCache data in the test json?

PR #776, opened 2019-04-18
allow memory swappiness to be set to 0

rdallman:
per: https://docs.docker.com/config/containers/resource_constraints/#--memory-swappiness-details

found when trying to set this value to 0, these bindings at present do not
allow it. the default behavior of this setting is to inherit the setting from
the host. in order to keep the default behavior of not sending in this value,
we need to allow sending in 'null' as well as 0, this is a fabled go issue and
omitempty with null is the way to accomplish this. this seems preferable to
not having an API change, removing omitempty and defaulting to turning this
off for everyone, as this is not the intended default. however, this is an API
change, though I suspect nobody tried to set it to 0 yet ;)

I've tested and this fixes the issue for me.

thank you!


rdallman:
let me know if this is my fault on CI, I've taken a glance and it just looks like some mirrors are down?

fsouza:
Hey @rdallman, thanks for contributing! The change looks good, but I wonder if we could get this done without a breaking change (though it's a fair breaking change as it's fixing a bug in the package api). I'll give this some thought and try a couple of things.

rdallman:
🤞 thank you!

PR #782, opened 2019-05-30
add DriverOpts to EndpointConfig

kppk:
Docker API v 1.32 introduced `DriverOpts` to `EndpointConfig`, which is defined as:

>DriverOpts is a mapping of driver options and values. These options are passed directly to the driver and are driver specific.

see https://docs.docker.com/engine/api/v1.32/#operation/ContainerCreate

This PR adds `DriverOpts` field to `EndpointConfig` struct.


fsouza:
Hi @kppk, thanks for contributing! Change looks good, let me investigate what's going on with Travis so we can get this in :)

fsouza:
It looks like it was an issue with staticcheck and Go 1.11, fixed in https://github.com/dominikh/go-tools/commit/a1efa522b896190dac280da996652c3394f89ff7. I'm restarting the builds and will merge once it's green.

Thanks for contributing!

PR #794, opened 2019-08-19
reorder config files

abrechon:
**rationale** there's still an empty `auths` json block in `$HOME/.docker/config.json` in the newest docker version therefore `plaintext-passwords.json` needs to be parsed first otherwise it returns an empty auth.


abrechon:
Hi @fsouza, can you please review and merge this one as well ! thanks 😸

PR #795, opened 2019-08-22
PullImageOptions add Platform argument per API 1.34

grooverdan:
Per https://docs.docker.com/engine/api/v1.34/#operation/ImageCreate I've added the Platform argument.

I've put the version into the URL which was added in [v1.18](https://docs.docker.com/engine/api/v1.18/#section/Versioning). [v1.26](https://docs.docker.com/engine/api/v1.26/#section/Versioning) was when accessing URLs without versions was deprecated however per current 1.40 hasn't removed it.

To be keep the change minimal adding the version only when Platform is specified.

Debugging with print statements verified the URL was populated only if a non-empty Platform was specified.

While https://github.com/fsouza/go-dockerclient-integration/pull/2 didn't fortunately update the test image to provide more concrete tests for this I hope this is acceptable.

If you'd like version tests/implementation to be different I'm happy to to change this.


image.go
```diff
@@ -318,6 +319,9 @@ func (c *Client) PullImage(opts PullImageOptions, auth AuthConfiguration) error
 		opts.Repository = parts[0]
 		opts.Tag = parts[1]
 	}
+	if len(opts.Platform) > 0 && (c.requestedAPIVersion == nil || c.requestedAPIVersion.LessThan(apiVersion134)) {
+		c.requestedAPIVersion = apiVersion134
```
fsouza: Is this really needed? 🤔 I wonder if we could get away without doing it. This creates the potential for a data race and I'd rather not introduce any synchronization here (currently requestedAPIVersion is only written during initialization).

grooverdan: Its not needed. I haven't tried to enter a platform on an api <1.34 server.  I looked at the docker code base on API 1.18 and every tried entering garbage params on whatever dockerhub has and they seem to be ignored.

So sure, can drop it.

Do you consider it a client library responsibility to check for a version number (requested and/or server) to conform with the API? Without that a user won't get an error. Putting in the version into the URL was my way of trying to get the server to check/report errors.

fsouza: @grooverdan ``c.requestedAPIVersion`` is being read pretty much by every method, so that's why it's written only initialization. The data race happens whenever any other goroutine reads that variable while PullImage modifies it.

I'd rather return an error here, instructing the user to set the required version, like we do with exec:

https://github.com/fsouza/go-dockerclient/blob/f3ce87165b89b84b35dd8d9a898e4f91c562cc23/exec.go#L46-L51

grooverdan: Added version test, though SkipServerVersionCheck is true normally.

Tricker bit was always adding the version to the URL if Platform was specified. Did this by adding `ver` tag to structures so every enhancement will explicitly include the minimum API version (unless requestedAPI is greater).

As such there's much more code added that expected.

fsouza: hmm I still think this is more than what we need, we can probably just give the power to the caller here, and as long as we report intuitive errors, the caller will know what needs to happen. Re: SkipServerVersionCheck being mostly true, I just opened #797, and I believe that that combined with your change in #796 should get us where we want to. What do you think?

grooverdan: I think its `go-dockerclient` responsibility to implement the API and use the minimum correct version in the URL as required for maximum compatibility. `go-dockerclient` has a clear structures and go functions that allows a user to avoid the need to work in knowing detailed API versions. If they need a later feature then it will have some API version constraints that `go-dockerclient` can know use the required version in the URL. By taking this approach we leave it to the server to report the error if that API version isn't supported.

The server side errors are quite clear on versioning without needing to implement them on the client side. An example API error will look like ```integration_test.go:100: API error (400): client version 1.44 is too new. Maximum supported API version is 1.40```. The mapping of this to a URL isn't done explicitly however it wouldn't take much in the `client.go` to pretend this to an error message if desired.

Checking against the server version, either within the library or the users of the library, is quite crude when `go-dockerclient` can just use the documented version and the server can complain if its that API version is not implemented. For these reasons I think #797 is the wrong approach and adds overhead in network round trips and complexity in the user's application needing to understand the API as well as the `go-dockerclient` interfaces (and how these relate to the API) and requires significant understanding to use these consistently and correctly. It isn't power we're giving the user, its the requirement to set the API version that the `go-dockerclient` needs to use the API interfaces that `go-dockerclient` implements.

Its for this reason that this PR adds versioned mechanics internal to `go-dockerclient` that allows features to be added and the URLs gets the version string while preserving the legacy [deprecated behaviour](https://docs.docker.com/engine/api/v1.26/#section/Versioning) of omitting versions. I've respected the legacy expectation that a user specified later version is used if specified, and errors if the newer features contradict that. #796 was added to preserve compatibility with other implementations however I still consider the user exposure to specify versions as legacy thing to be avoided.

grooverdan:
rebased to remove fixmie.

Also found that the API version limit is 1.32 not 1.34 (lodged docker bug https://github.com/docker/docker.github.io/issues/9305)

client.go
```diff
@@ -897,53 +933,80 @@ func queryString(opts interface{}) string {
...
 	case reflect.Map:
 		if len(v.MapKeys()) > 0 {
 			if b, err := json.Marshal(v.Interface()); err == nil {
 				items.Add(key, string(b))
+				return true
 			}
 		}
 	case reflect.Array, reflect.Slice:
 		vLen := v.Len()
+		var values_added int
```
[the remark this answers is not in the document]
grooverdan: sure. Apologies, was a quick test to avoid what might be a golang bug that I didn't clean up.

grooverdan:
Thanks for the quick review/merge.

PR #799, opened 2019-08-28
BuildImage updated to v1.40 API

grooverdan:
added to BuildImageOptions:
* ExtraHosts
* ShmSize
* Platform
* Outputs

Removed `qs` tags from options where the parameter URL was the lower
case of the param name (as this is the default).

Added `ver:`sion tags on all BuildImageOptions based on docker
server code api/server/router/build/build_routes.go:postBuild,
https://docs.docker.com/engine/api/v1.40/#operation/ImageBuild, and
https://docs.docker.com/engine/api/version-history/.

`forcerm`, `pull` and `output` parameters are the only version checked in the
docker server code, however applying Postel's Law we'll be explicit.

Not included is the experimental option "squash" and the undocumented
"isolation" option.

I wouldn't rush to do yet another release - I'm looking at the container APIs based on openshift/imagebuilder and containers/buildah usage.


grooverdan:
Thanks for the quick merge.

PR #807, opened 2019-10-08
Fix plugin installation and removal on latest docker versions

stffabi:


stffabi:
Thanks so much for the merge. I have another PR in the pipeline which contains some more fixes for the plugin api...

PR #810, opened 2019-11-05
Support customizing ShmSize parameter when creating container

Soulou:
https://docs.docker.com/engine/api/v1.30/#operation/ContainerCreate


Soulou:
My bad it is already present in `HostConfig` my apologies

PR #820, opened 2020-02-03
base64 string without padding on docker for mac

abrechon:
**rationale** On docker for mac, the base64 string stored in `plaintext-passwords.json` has no padding. plus adds a clearer error when base64 decoding fails

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -170,9 +170,14 @@ func authConfigs(confs map[string]dockerConfig) (*AuthConfigurations, error) {
 		if conf.Auth == "" {
 			continue
 		}
+
+		// support both padded and unpadded encoding
 		data, err := base64.StdEncoding.DecodeString(conf.Auth)
 		if err != nil {
-			return nil, err
+			data, err = base64.StdEncoding.WithPadding(base64.NoPadding).DecodeString(conf.Auth)
+		}
+		if err != nil {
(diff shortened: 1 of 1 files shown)


abrechon:
missing tests update, on its way.

abrechon:
> Hi @abrechon, thanks for contributing. I don't know if this is the right fix, just in case people are using some weird version of docker on darwin that's not docker for mac (or they manage the version of the docker cli separately). Perhaps we could support both? (use base64.RawStdEncoding if `len(conf.Auth) % 4 != 0`?)

Hi @fsouza , ok what about this version that implements both decoding methods ?

fsouza:
@abrechon that's better! 😁

auth.go
```diff
@@ -170,9 +170,14 @@ func authConfigs(confs map[string]dockerConfig) (*AuthConfigurations, error) {
 		if conf.Auth == "" {
 			continue
 		}
+
+		// support both padded and unpadded encoding
 		data, err := base64.StdEncoding.DecodeString(conf.Auth)
 		if err != nil {
-			return nil, err
+			data, err = base64.StdEncoding.WithPadding(base64.NoPadding).DecodeString(conf.Auth)
```
fsouza:
```suggestion
			data, err = base64.RawStdEncoding.DecodeString(conf.Auth)
```

PR #832, opened 2020-05-18
Container operations divided separate files

UmutComlekci:
For better reading and development, `container.go` file divided separate files


fsouza:
Nice thanks for taking care of this, I'll have a look later in the week, but in the meantime can you split container_test.go too?

fsouza:
Thank you very much! I'll look into this over the weekend.

PR #847, opened 2020-11-01
Add the possiblity to call docker credentials helpers

teintuc:
I added the possibility to call docker credentials helpers.
It will reads from docker config files to try to find providers and then run an installed credentials helpers .

I'm opened to suggestion to make it better (specially advice to make good tests) :)

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -360,14 +360,15 @@ func getCredentialsFromHelper(provider string, registry string) (*helperCredenti
 	}
 
 	c := new(helperCredentials)
-	if err := json.NewDecoder(helpercreds).Decode(c); err != nil {
+	err = json.Unmarshal(helpercreds, c)
+	if err != nil {
 		return nil, err
 	}
 
 	return c, nil
 }
 
-func runDockerCredentialsHelper(provider string, registry string) (*bytes.Buffer, error) {
+func runDockerCredentialsHelper(provider string, registry string) ([]byte, error) {
 	cmd := exec.Command("docker-credential-"+provider, "get") //nolint:gosec
 
 	var stdout bytes.Buffer
@@ -380,5 +381,5 @@ func runDockerCredentialsHelper(provider string, registry string) (*bytes.Buffer
 		return nil, err
 	}
 
-	return &stdout, err
+	return stdout.Bytes(), nil
 }
--- a/auth.go
+++ b/auth.go
@@ -368,7 +368,7 @@ func getCredentialsFromHelper(provider string, registry string) (*helperCredenti
 }
 
 func runDockerCredentialsHelper(provider string, registry string) (*bytes.Buffer, error) {
-	cmd := exec.Command("docker-credential-"+provider, "get")
+	cmd := exec.Command("docker-credential-"+provider, "get") //nolint:gosec
 
 	var stdout bytes.Buffer
 
--- a/auth.go
+++ b/auth.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"path"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


teintuc:
I add the linter raising the exec line. After some digging, I add to ignore the line. Not sur if it is the best solution

auth.go
```diff
@@ -283,3 +284,101 @@ func (c *Client) AuthCheck(conf *AuthConfiguration) (AuthStatus, error) {
...
+type helperCredentials struct {
+	Username string `json:"Username,omitempty"`
+	Secret   string `json:"Secret,omitempty"`
+}
+
+// NewAuthConfigurationsFromCredsHelpers returns AuthConfigurations from
+// installed credentials helpers
+func NewAuthConfigurationsFromCredsHelpers(registry string) (*AuthConfiguration, error) {
+	// Load docker configuration file in order to find a possible helper provider
+	pathsToTry := cfgPaths(os.Getenv("DOCKER_CONFIG"), os.Getenv("HOME"))
```
fsouza, line 298: Shouldn't this be part of `NewAuthConfigurationsFromDockerCfg`?

Also, can you add a test?

auth.go
```diff
@@ -283,3 +284,101 @@ func (c *Client) AuthCheck(conf *AuthConfiguration) (AuthStatus, error) {
...
+	c := new(helperCredentials)
+	if err := json.NewDecoder(helpercreds).Decode(c); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func runDockerCredentialsHelper(provider string, registry string) (*bytes.Buffer, error) {
+	cmd := exec.Command("docker-credential-"+provider, "get") //nolint:gosec
```
fsouza, line 371: Yeah gosec will complain about us executing a user provided string, but this should be fine since we're constraining the commanad that gets executed.

auth.go
```diff
@@ -283,3 +284,101 @@ func (c *Client) AuthCheck(conf *AuthConfiguration) (AuthStatus, error) {
...
+
+	c := new(helperCredentials)
+	if err := json.NewDecoder(helpercreds).Decode(c); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func runDockerCredentialsHelper(provider string, registry string) (*bytes.Buffer, error) {
```
fsouza, line 370: Can we return just []byte or io.Reader here? Just for a better abstraction (not extremely important though as this is an unexported function).

auth.go
fsouza: (for the test, feel free to add a script to testdata and invoke that, it's easier than mocking exec.Command)

auth.go
teintuc: ok. So I'm going to leave it like this then

auth.go
teintuc: At first, I wanted to put as a second way of authenticating in `NewAuthConfigurationsFromDockerCfg` but the function name clearly state it comes from docker config. Which not really what I'm doing. Thats why I made this like this. 

I have no problem doing it though. It's just me overthinking it :)

I will add a test.

auth.go
teintuc: I will change it. At least, someone will be able to use it for something else more easily.
As you said, not really important, but I'll follow your directions

auth.go
teintuc: After some thinking, I'll need your opinion.

I also did it because I need to pass the registry name in order to feed it to the found helper. If I put it in `NewAuthConfigurationsFromDockerCfg`, it would change the function way (thinking about backward compatibility).

What do you think?

teintuc: Sorry for pushing but could you give me your opinion about my last comment ? Thanks

fsouza: Hey @teintuc, sorry for taking so long to get back to you. Things have been crazy over here x)

> After some thinking, I'll need your opinion.
> 
> I also did it because I need to pass the registry name in order to feed it to the found helper. If I put it in `NewAuthConfigurationsFromDockerCfg`, it would change the function way (thinking about backward compatibility).
> 
> What do you think?

This is a good point that I didn't realize in my original message. We can keep this version, as we need to take the registry. Let me just give this another look before merging! 

Thank you very much for your contribution and patience!

fsouza (approved):
Thank you very much for contributing!

PR #854, opened 2020-11-26
Ports to string

stokito:
The PR adds two methods to format port bindings into one string i.e. the same as docker ps shows.
Originally the function was written in [ctop](https://github.com/bcicen/ctop/blob/65e9c6dff6a4ac5922c5c18e4aba13142a802efa/connector/docker.go#L133) but it looks like it should be moved to docker client.
The ctop license is MIT and it looks like we can copy that code to docker client.

### DIFF
--- a/container.go
+++ b/container.go
@@ -524,78 +524,78 @@ type BlockLimit struct {
 // HostConfig contains the container options related to starting a container on
 // a given host
 type HostConfig struct {
-	Binds                []string               `json:"Binds,omitempty" yaml:"Binds,omitempty" toml:"Binds,omitempty"`
-	CapAdd               []string               `json:"CapAdd,omitempty" yaml:"CapAdd,omitempty" toml:"CapAdd,omitempty"`
-	CapDrop              []string               `json:"CapDrop,omitempty" yaml:"CapDrop,omitempty" toml:"CapDrop,omitempty"`
-	Capabilities         []string               `json:"Capabilities,omitempty" yaml:"Capabilities,omitempty" toml:"Capabilities,omitempty"` // Mutually exclusive w.r.t. CapAdd and CapDrop API v1.40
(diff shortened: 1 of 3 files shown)


stokito:
It will be better to rename PortBindings to PortMap. Also It looks like my IDE changed line endings. I'll fix and resend

fsouza (commented):
Hey @stokito, thanks for contributing!

The codei looks good overall, but unfortunately we can't merge it as is, since it would introduce a breaking change in the client. We could provide helper functions for users of the client, but I don't know if that's something that belongs here? What do you think?

stokito:
Well, the new type is just an alias. When I used it even tests still was compilable. I just changed tests to make better usage examples.
I'm new to Go but as far I understood this change wont break compilation. Or it can?
Anyway, we can increment major version.

fsouza:
Hey, apologies for the delayed response.

The type is not an alias and the tests compiled because the test is only consuming the API, so yeah the serialization format didn't change, but the actual Go representation changed. A test/code  that did something like this wouldn't compile:

```go
portBindings := map[Port][]PortBinding{...}
client.CreateContainer(docker.CreateContainerOptions{
    HostConfig: &docker.HostConfig{
        ...
        PortBindings: portBindings,
    },
})
```

That's why you had to make these changes: https://github.com/fsouza/go-dockerclient/pull/854/files#diff-6d06f36cb77e9d02e6d4083cdd4791b244bd93aa106b5e7c30035fc4c8ac1156L495-R495

I really appreciate the cleanup, but introducing a new major release because of such changes would be painful. Major releases with Go modules are painful and I'd prefer to have a better planned refactoring to fix some other things that haven't been fixed due to backwards compatibility promises, but then there's no real incentive to actually work on those things, since Docker has an official SDK nowadays.

We could provide these methods as new functions in the package, but I'm not sure that even makes sense? Is this something you saw a lot in other packages besides ctop? Or something you  need for a different project?

stokito:
Thank you. I don’t know how to check other apps if they have similar port format function. 
Then let’s close the pr

PR #855, opened 2020-11-27
Args type for filters

stokito:
The Args type is similar to [filters.Args](https://github.com/moby/moby/blob/c9b2a3cff51a6403f4eac319af5a794b0dbdeba4/api/types/filters/parse.go) type from official Docker client.
It also provides constructor NewArgs() and method Add().
This can make migration to official docker client easier and reuse familiar concept


fsouza (commented):
Same as #854, while this makes the code look better, it introduces a breaking change to users of the library :(

stokito:
Closing this because author decided to avoid breaking changes

PR #856, opened 2020-11-27
Events filters

stokito:
First three commits are from another PR #855
The PR add filtering to events and this is very important because there is a lot of events especially if containers have health checks.
I created EventsOptions but at this moment I have no idea how to support `since` and `until` criterias so I commented them out.

### DIFF
--- a/event.go
+++ b/event.go
@@ -7,12 +7,12 @@ package docker
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"math"
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


fsouza:
Sorry I missed this. Will take a look tomorrow.

stokito:
You can just cherry pick the last commit

stokito:
Now the PR is ready to be merged.
I removed those commits with new Args type: instead used just plain `Filters map[string][]string`.
I kept logic with Since

fsouza (commented):
One minor comment and a question about the proper API. Taking time would allow users to do something like:

```go
eventsOptions := docker.EventsOptions{
    Since:  time.Now().Add(-2 * time.Hour),
    Filters: map[string][]string{...},
}
```

Alternatively we could just take int64 which would be the unix timestamp (though that would make it difficult to differentiate between zero and unset).

stokito:
Totally agree that time would be better than string. I keep it as a string for two reasons:
1. In other places we already using string for the `Since` field https://github.com/fsouza/go-dockerclient/blob/214a51d9a1e5c9cb0c69641f326c2d835820f35b/container_list.go#L16
2. Official docker client uses string and it can be a Duration or formatted RFC3339 date:
    * https://github.com/moby/moby/blob/5526afa2415bd3ee9b6878378247a9e07e44a7ef/api/types/client.go#L105
    * https://github.com/moby/moby/blob/master/client/events.go#L77
    * https://github.com/moby/moby/blob/5526afa2415bd3ee9b6878378247a9e07e44a7ef/api/types/time/timestamp.go#L26

Even while now we support only timestamp you may want to change this in future.

fsouza (approved):
Thanks for contributing!

PR #864, opened 2021-03-04
ServerAPIVersion needs to be set in Client.CreateExec()

rfay:
`c.checkAPIVersion()` may not have been called before the client is used in CreateExec(), and thus there's no ServerApiVersion, so the version checks go all wrong.


fsouza (approved):
Thanks for contributing!

PR #922, opened 2022-06-01
add fields definition for DockerStat API of Docker host uses cgroup2

### LINKED ISSUE #919: Missing some fields to support Stats API on Docker host uses cgroup2
Docker Engine 20.10.0+ supports cgroup2 and uses it by default.
> Support cgroup2 moby/moby#40174 moby/moby#40657 moby/moby#40662
> cgroup2: use “systemd” cgroup driver by default when available moby/moby#40846

(cf. https://docs.docker.com/engine/release-notes/#20100 )

The output of Docker Stats API varies depending on the version of cgroup used by Docker Engine.
Especially, `Stats.MemoryStats.Stats` for cgroup2 host much differ from the one of cgroup.

(definition for cgroup v1)
https://github.com/moby/moby/blob/548614694305d10e25665b4a7d0f61b025d47dd1/daemon/daemon_unix.go#L1439-L1470

(definition for cgroup v2)
https://github.com/moby/moby/blob/548614694305d10e25665b4a7d0f61b025d47dd1/daemon/daemon_unix.go#L1544-L1577

Currently, `Stats` (https://github.com/fsouza/go-dockerclient/blob/main/container_stats.go#L26) struct has only fields for cgroup v1, so it would be better to add the fields for cgroup v2 as well.
https://github.com/fsouza/go-dockerclient/blob/main/container_stats.go#L26

xruins:
This PR adds some fields to support Docker Stat API of Docker host uses cgroup2.
In addition, added test codes for Docker Stat API with cgroup2.

fix #919

### DIFF
--- a/container_stats.go
+++ b/container_stats.go
@@ -55,6 +55,30 @@ type Stats struct {
 			TotalPgpgin             uint64 `json:"total_pgpgin,omitempty" yaml:"total_pgpgin,omitempty" toml:"total_pgpgin,omitempty"`
 			HierarchicalMemswLimit  uint64 `json:"hierarchical_memsw_limit,omitempty" yaml:"hierarchical_memsw_limit,omitempty" toml:"hierarchical_memsw_limit,omitempty"`
 			Swap                    uint64 `json:"swap,omitempty" yaml:"swap,omitempty" toml:"swap,omitempty"`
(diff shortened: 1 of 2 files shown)


xruins:
You’re welcome. Thanks you for quick review.

PR #947, opened 2022-12-12
fix(container): The response body will not be closed after an error

guowei-gong:


fsouza (commented):
Can you clarify the issue you're facing? With your change, the code is likely to cause a nil pointer dereference, because whenever err is not nil, resp will be nil.

guowei-gong:
> Can you clarify the issue you're facing? With your change, the code is likely to cause a nil pointer dereference, because whenever err is not nil, resp will be nil.

Sorry, I didn't take into account the scenario you said, modifying it just because I think `body` should always be closed

fsouza:
Gotcha. In general, the convention in Go is to assume that if there's an error, there's no response, therefore no body to be closed. Unfortunately, this can't be guaranteed by the type system, but we are using a package from the standard library to send requests, so we should be safe.

Thanks for contributing!

guowei-gong:
> 

tks u

PR #961, opened 2023-02-07
Add the `Version` field to `BuildImageOptions`

nalind:
The `Version` field was added in API 1.39, though at the time the only known value was "1" ("classic" docker build).  API 1.42 added "2" to identify the newer BuildKit-based builder.
We don't bump the requested API version to 1.42 when we're passed "2", but we could, though I suspect that it was accepted but undocumented for some versions in between.

### DIFF
--- a/image.go
+++ b/image.go
@@ -469,6 +469,14 @@ func (c *Client) ImportImage(opts ImportImageOptions) error {
 	return c.createImage(&opts, nil, opts.InputStream, opts.OutputStream, opts.RawJSONStream, opts.InactivityTimeout, opts.Context)
 }
 
+// BuilderVersion represents either the BuildKit or V1 ("classic") builder.
+type BuilderVersion string
+
+const (
+	BuilderV1       BuilderVersion = "1"
+	BuilderBuildKit BuilderVersion = "2"
+)
+
 // BuildImageOptions present the set of informations available for building an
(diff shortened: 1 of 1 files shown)


image.go
```diff
@@ -469,6 +469,14 @@ func (c *Client) ImportImage(opts ImportImageOptions) error {
 	return c.createImage(&opts, nil, opts.InputStream, opts.OutputStream, opts.RawJSONStream, opts.InactivityTimeout, opts.Context)
 }
 
+// BuilderVersion represents either the BuildKit or V1 ("classic") builder.
+type BuilderVersion string
+
+const (
+	BuilderV1       BuilderVersion = "1"
```
fsouza, line 476: should this be `BuilderClassic`?

fsouza (commented):
Thanks for contributing!

image.go
nalind: The client SDK ([here](https://github.com/moby/moby/blob/5f369ff1d4c8cc2162f865d11f2fbcd5d612013d/api/types/client.go#L221)) calls it BuilderV1, but I've no real preference.  Should I change it?

fsouza: Let's keep it V1 then. Thanks for checking!

fsouza (approved):
Thank you!

nalind:
Thanks for the quick review!

Issue #994, opened 2023-07-02
Wrong error propagation in context


liran-funaro:
When the context is canceled, we should propagate the cause of the cancellation, not the error itself.
The current behavior would propagate "context canceled" as the cancel cause, which only makes sense if the context was canceled without an error. Otherwise, the actual cause should be used.

Fortunately, `context.Cause(ctx)` implements precisely this logic.

The following:
https://github.com/fsouza/go-dockerclient/blob/580dd4994ec2ffa0b02ba3823936e965d2a1dfc3/client.go#L511-L518

Should be replaced with:
```go
 func chooseError(ctx context.Context, err error) error { 
 	select { 
 	case <-ctx.Done(): 
 		return context.Cause(ctx)
 	default: 
 		return err 
 	} 
 } 
```

bencourliss:
So it looks like backwards compatibility with Go 1.19 was broken in a patch release of this library? https://github.com/fsouza/go-dockerclient/pull/998#issuecomment-1625593131

fsouza:
> So it looks like backwards compatibility with Go 1.19 was broken in a patch release of this library? https://github.com/fsouza/go-dockerclient/pull/998#issuecomment-1625593131

Yes, once Go 1.21 is out, Go 1.19 is no longer supported and we don't have to keep compatibility with it. If you need to keep using Go 1.19, don't upgrade go-dockerclient.

PR #995, opened 2023-07-02
Fix wrong error propagation in context

liran-funaro:
See issue #994


fsouza (approved):
Now that Go 1.21 is out, we can drop Go 1.19 and ship this.

PR #998, opened 2023-07-07
Fix wrong error propagation in context

fsouza:
Reapplying #995.

@liran-funaro I'm reopening this as the change looks good. Would you rather have me do it instead?

### DIFF
--- a/client.go
+++ b/client.go
@@ -511,7 +511,7 @@ type streamOptions struct {
 func chooseError(ctx context.Context, err error) error {
 	select {
 	case <-ctx.Done():
-		return ctx.Err()
+		return context.Cause(ctx)
 	default:
 		return err
 	}


fsouza:
(we can leave this open until Go 1.21 is released, then we can remove support for 1.19)

liran-funaro:
I closed this because after looking further, it seems correct to returns the "context closed" reason as this IS the reason you stopped the operating.
The user should investigate further on its own and find out why the context was closed.
The user own the context so it makes the most sense.
This is also the common practice and how it is implemented in other libraries.

fsouza:
> This is also the common practice and how it is implemented in other libraries.

Yeah, but Cause is new in Go 1.20, so I imagine adoption will lag.

liran-funaro:
I reopened #994 #995

But I suggest looking further into golang documentation do find out if they have suggestions on this matter.

I'll update on #994 if I'll find anything.

PR #1031, opened 2024-01-28
[testing/server] return container labels when listing containers

taraspos:
Testing server behavior is currently different from the actual server when listing containers.

When listing containers on a testing server labels map is always `nil` and the additional call is needed to `inspectContainer`, however, this is not the case on the actual docker server.

This PR sets labels for the listed container to fix that.


fsouza (approved):

fsouza:
Thanks for contributing!

PR #1083, opened 2024-08-20
add some log print event lost

ningmingxiao:


bufdev:
@ningmingxiao I have not been involved in this repo for about a decade, it's not really professional to tag people here. Will leave this to others.

ningmingxiao:
sory,but thank you anyway.

fsouza:
Thanks for contributing! This is a big change that can lead to slow consumers hanging. I recommend either making sure your caller is always fast enough to consume the events, or putting the new behavior behind a knob.

ningmingxiao:
sometimes I can make sure consumers run faster, if run plenty of container,  will generate many events, channel will be fulled. At least we should let user know event is dropped.

fsouza:
Can you add a knob in that case?

ningmingxiao:
> Can you add a knob in that case?

Could you describe it in more detail? Thank you

fsouza:
@ningmingxiao I imagine we'd have a new option when adding the listener. Internally the listener would become some struct with the channel + some listening options, and externally we'd probably need to have versions of `AddEventListenerWithOptions` and `RemoveEventListener` that take a struct with "listening options" instead of just the channel.

PR #1183, opened 2026-05-18
ImportImageOptions: add Platform field for cross-arch image import

paperbenni:
PullImageOptions and BuildImageOptions already expose this field.

I need this to add multi-arch support to https://github.com/involucro/involucro and by extension mulled and bioconda. The lack of this currently causes unnecessary caveat when it comes to supporting different CPU architectures.


fsouza (approved):
Thanks for contributing!

PR #1186, opened 2026-06-04
Add docker CLI contexts support to NewClientFromEnv

ebriney:
## Summary
- New `NewContextClient(name)` / `NewVersionedContextClient(name, apiVersion)` build a client from a named docker CLI context by reading `$DOCKER_CONFIG/contexts/meta/<sha256(name)>/meta.json` and the optional TLS bundle under `contexts/tls/<id>/docker/`. The reserved name `default` (or `""`) resolves to the platform default host.
- `NewClientFromEnv` / `NewVersionedClientFromEnv` now match the docker CLI's resolution order: `DOCKER_HOST` keeps full precedence; when it's unset the client falls back to `DOCKER_CONTEXT`, then the `currentContext` field of `config.json`, then the default socket.
- `ssh://` context endpoints are rejected with a clear error since this library has no SSH dialer.

## Test plan
- [x] `go test ./...` (unit tests, Linux/macOS)
- [x] Manual smoke test on Windows
- [ ] `NewClientFromEnv` with `DOCKER_HOST` set still hits that host (existing tests cover this)
- [ ] `NewClientFromEnv` with a `docker context use <name>` selecting a `tcp://` context picks up that host
- [ ] `NewContextClient("default")` resolves to the platform default host without touching disk
- [ ] `NewContextClient` on a missing context returns a clear "not found" error
- [ ] `NewContextClient` on an `ssh://` context returns an explanatory error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/client.go
+++ b/client.go
@@ -262,6 +262,10 @@ func NewVersionedTLSClient(endpoint string, cert, key, ca, apiVersionString stri
 // Docker's default logic for the environment variables DOCKER_HOST, DOCKER_TLS_VERIFY, DOCKER_CERT_PATH,
 // and DOCKER_API_VERSION.
 //
+// When DOCKER_HOST is unset, the host is resolved from the active docker CLI
+// context (DOCKER_CONTEXT, or the currentContext field of
+// $DOCKER_CONFIG/config.json), matching the docker CLI's behavior.
+//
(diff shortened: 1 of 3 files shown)


fsouza:
Sorry I lost track of this, can you rebase so CI can run again?

PR #1187, opened 2026-06-05
Improve API version handling

### LINKED ISSUE #1030: With the Latest Docker Engine v25.x the Version Selection Magic Based on the Call Option 'ver` Tags Results in Failing Calls
The Docker Engine v25.0 release depricates the APIs below version 1.24 ( https://docs.docker.com/engine/deprecated/#deprecate-legacy-api-versions ).

The current dynamic version selection logic causes failures if you haven't configured an options field with a big enough 'ver' tag.

We can use `(c *Client) BuildImage(opts BuildImageOptions)` as an example: https://github.com/fsouza/go-dockerclient/blob/main/image.go#L537

It calls `queryStringVersion` to figure out the minimal API version based on the provided `BuildImageOptions` fields. Then it calls `pathVersionCheck`, which creates `buildURL` based either on the explicitly configured API version when the client was created or the minimal API version guessed by `queryStringVersion`.

If you set the `Name` and `ForceRmTmpContainer` fields in `BuildImageOptions` `queryStringVersion` will guess the minimal version to be `1.12`. Then `pathVersionCheck` will create an API path with `1.12` if you don't create a versioned Client explicitly.

The go-dockerclient claims that if you don't specify an API version then the latest API version will be used, but it's not really the case here.

@fsouza curious what you think about this behavior and how you'd prefer to address it in the library. The docker/moby client has the `NegotiateAPIVersion` call ( https://github.com/moby/moby/blob/v25.0.1/client/client.go#L294 ) that's potentially relevant for this.

paperbenni:
The client requests the lowest API version containing all the features used in a request. This used to just provide support for older docker versions if newer features simply weren't used. It causes the client to be rejeceted by new docker versions because old API versions are explicitly no longer supported. This changes API version negotiation to be compatible with newer docker versions

This makes versioning negotiation somewhat more complex, requiring the other changes to make thread safe. 
The race conditions were technically preexisting but rarely triggered because checking the server version was skipped most of the time

Closes #1030.

### DIFF
--- a/client.go
+++ b/client.go
@@ -146,9 +146,11 @@ type Client struct {
 	endpointURL         *url.URL
 	eventMonitor        *eventMonitoringState
 	requestedAPIVersion APIVersion
-	serverAPIVersion    atomic.Value
-	expectedAPIVersion  atomic.Value
-	versionMu           sync.Mutex
+	// serverAPIVersion and expectedAPIVersion are read while requests are built,
+	// so keep access lock-free after the first version probe completes.
+	serverAPIVersion   atomic.Pointer[APIVersion]
+	expectedAPIVersion atomic.Pointer[APIVersion]
+	versionMu          sync.Mutex
 }
 
 // Dialer is an interface that allows network connections to be dialed
@@ -365,7 +367,7 @@ func (c *Client) getExpectedVersion() APIVersion {
 	if v == nil {
 		return nil
 	}
-	return v.(APIVersion)
+	return *v
 }
 
 func (c *Client) getServerVersion() APIVersion {
@@ -373,7 +375,11 @@ func (c *Client) getServerVersion() APIVersion {
 	if v == nil {
 		return nil
 	}
-	return v.(APIVersion)
+	return *v
+}
+
+func storeAPIVersion(p *atomic.Pointer[APIVersion], v APIVersion) {
+	p.Store(&v)
 }
 
 func (c *Client) ensureAPIVersion() error {
@@ -383,14 +389,16 @@ func (c *Client) ensureAPIVersion() error {
 	return c.ensureServerVersion()
 }
 
+// setExpectedVersion records the version used in request URLs. It may differ
+// from the server version when the caller requested an explicit API version.
 func (c *Client) setExpectedVersion(serverVersion APIVersion) {
 	if c.SkipServerVersionCheck || c.getExpectedVersion() != nil {
 		return
 	}
 	if c.requestedAPIVersion != nil {
-		c.expectedAPIVersion.Store(c.requestedAPIVersion)
+		storeAPIVersion(&c.expectedAPIVersion, c.requestedAPIVersion)
 	} else {
-		c.expectedAPIVersion.Store(serverVersion)
+		storeAPIVersion(&c.expectedAPIVersion, serverVersion)
 	}
 }
 
@@ -413,11 +421,14 @@ func (c *Client) ensureServerVersion() error {
 	if err != nil {
 		return err
 	}
-	c.serverAPIVersion.Store(serverAPIVersion)
+	storeAPIVersion(&c.serverAPIVersion, serverAPIVersion)
 	c.setExpectedVersion(serverAPIVersion)
 	return nil
 }
 
+// ensureServerVersionForCompatibility probes /version only to decide old API
+// compatibility behavior. Respect SkipServerVersionCheck by preserving the old
+// best-effort semantics when that probe fails.
 func (c *Client) ensureServerVersionForCompatibility() error {
 	err := c.ensureServerVersion()
 	if c.SkipServerVersionCheck {
--- a/client.go
+++ b/client.go
@@ -403,6 +403,8 @@ func (c *Client) ensureServerVersion() error {
 		c.setExpectedVersion(serverVersion)
 		return nil
 	}
+	// Serialize only the first /version request; version reads stay lock-free
+	// once the server version has been cached.
 	c.versionMu.Lock()
 	defer c.versionMu.Unlock()
 	if serverVersion := c.serverAPIVersion.Load(); serverVersion != nil {
@@ -420,17 +422,6 @@ func (c *Client) ensureServerVersion() error {
 	c.serverAPIVersion.Store(serverAPIVersion)
 	c.setExpectedVersion(serverAPIVersion)
 	return nil
-}
-
-// ensureServerVersionForCompatibility probes /version only to decide old API
-// compatibility behavior. Respect SkipServerVersionCheck by preserving the old
-// best-effort semantics when that probe fails.
-func (c *Client) ensureServerVersionForCompatibility() error {
-	err := c.ensureServerVersion()
-	if c.SkipServerVersionCheck {
-		return nil
-	}
-	return err
 }
 
 // Endpoint returns the current endpoint. It's useful for getting the endpoint
@@ -946,6 +937,8 @@ func (c *Client) pathVersionCheck(basepath, queryStr string, requiredAPIVersion 
 				basepath, requiredAPIVersion, expected)
 		}
 	}
+	// Return only a request path. The caller applies the endpoint and selected
+	// API version later through getURL.
 	return fmt.Sprintf("%s?%s", basepath, queryStr), nil
 }
 
--- a/client.go
+++ b/client.go
@@ -148,8 +148,8 @@ type Client struct {
 	requestedAPIVersion APIVersion
 	// serverAPIVersion and expectedAPIVersion are read while requests are built,
 	// so keep access lock-free after the first version probe completes.
-	serverAPIVersion   atomic.Pointer[APIVersion]
-	expectedAPIVersion atomic.Pointer[APIVersion]
+	serverAPIVersion   atomicAPIVersion
+	expectedAPIVersion atomicAPIVersion
 	versionMu          sync.Mutex
 }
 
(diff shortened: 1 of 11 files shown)


image.go
```diff
@@ -408,27 +412,36 @@ func (c *Client) ExportImages(opts ExportImagesOptions) error {
...
-	var exporturl string
-	if c.requestedAPIVersion.GreaterThanOrEqualTo(apiVersion125) {
+	var exportpath string
+	var effectiveVersion APIVersion
+	if expected := c.expectedAPIVersion.Load(); expected != nil {
+		effectiveVersion = expected
+	} else if c.requestedAPIVersion != nil {
+		effectiveVersion = c.requestedAPIVersion
+	}
+	if effectiveVersion != nil && effectiveVersion.GreaterThanOrEqualTo(apiVersion125) {
```
fsouza, line 426: This breaks multi-image export for negotiating clients. The comma-join format now applies whenever the negotiated server version is >= 1.25, but the daemon's `/images/get` handler reads `r.Form["names"]` without splitting on commas — the joined string goes straight to `reference.ParseAnyReference`, which rejects commas. So `ExportImages` with two names against a modern daemon, which worked before (repeated `names=` params), now errors. The "1.25 uses comma-separated" note traces back to a Swagger `collectionFormat` artifact, not actual daemon behavior. I'd drop the comma-join branch entirely and always send repeated `names=` params (note `TestExportImagesNegotiatedVersionFirstCall` currently locks in the broken wire format). Also, if the join does stay for some reason, `strings.Join(opts.Names, ",")` would replace the manual loop.

client.go
```diff
@@ -878,21 +921,25 @@ func (c *Client) getPath(basepath string, opts any) (string, error) {
...
+	if expected := c.expectedAPIVersion.Load(); expected != nil {
+		if requiredAPIVersion != nil && !expected.GreaterThanOrEqualTo(requiredAPIVersion) {
+			return "", fmt.Errorf("API %s requires version %s, server version %s is insufficient",
+				basepath, requiredAPIVersion, expected)
+		}
 	}
-	return fmt.Sprintf("%s%s?%s", urlStr, basepath, queryStr), nil
+	// Return only a request path. The caller applies the endpoint and selected
+	// API version later through getURL.
+	return fmt.Sprintf("%s?%s", basepath, queryStr), nil
```
fsouza, line 942: For `SkipServerVersionCheck` clients — which is the default for `NewClient`/`NewTLSClient` — this now neither pins the URL to `requiredAPIVersion` nor validates anything: `ensureAPIVersion` is a no-op under skip, `expectedAPIVersion` stays nil, and the request goes out unversioned. Before, `PullImage` with `Platform` (`ver:"1.32"`) against an old daemon was forced to `/v1.32/...` and failed loudly; now the old daemon silently ignores the unknown `platform` param and "successfully" pulls the wrong-architecture image. When skip is on and there's no expected version, I think the required version still needs to be either pinned into the URL (old behavior) or validated somehow, rather than dropped.

client.go
```diff
@@ -358,20 +362,65 @@ func (c *Client) SetTimeout(t time.Duration) {
...
+}
+
+func (c *Client) probeServerVersion() error {
+	if serverVersion := c.serverAPIVersion.Load(); serverVersion != nil {
+		c.setExpectedVersion(serverVersion)
+		return nil
+	}
+	// Serialize only the first /version request; version reads stay lock-free
+	// once the server version has been cached.
+	c.versionMu.Lock()
```
fsouza, line 408: `versionMu` is held across the `/version` HTTP round-trip, which runs on `context.Background()` (and `defaultClient()` sets no `http.Client.Timeout`). A daemon that accepts connections but stalls on `/version` hangs the first caller forever and parks every other `do()`/`stream()`/`hijack()` call — plus the unconditional probes in `StartContainer`/`CreateExec`/`CopyFromContainer` — on an uncancellable mutex. And since failures are never cached, a persistently failing `/version` turns N concurrent calls into N *serialized* probes. Consider doing the HTTP call outside the lock (e.g. singleflight-style, or accept a rare duplicate probe), threading the caller's context into the probe, and/or caching probe failure.

client.go
```diff
@@ -145,8 +146,11 @@ type Client struct {
 	endpointURL         *url.URL
 	eventMonitor        *eventMonitoringState
 	requestedAPIVersion APIVersion
-	serverAPIVersion    APIVersion
-	expectedAPIVersion  APIVersion
+	// serverAPIVersion and expectedAPIVersion are read while requests are built,
+	// so keep access lock-free after the first version probe completes.
+	serverAPIVersion   atomicAPIVersion
```
fsouza, line 151: Adding a `sync.Mutex` and `atomic.Pointer`-based fields makes the exported `Client` struct non-copyable: `c2 := *client` was vet-clean before and now trips `copylocks` (this PR itself had to change `newTestClient` from returning `Client` to `*Client`). Downstream code that copies `Client` by value will start failing vet-gated CI, and a runtime copy forks the version cache or carries a locked mutex. Maybe unavoidable for the thread-safety goal, but worth an explicit decision — and a mention in the changelog if we accept it.

image.go
```diff
@@ -408,27 +412,36 @@ func (c *Client) ExportImages(opts ExportImagesOptions) error {
...
 		})
 	} else {
-		exporturl, err = c.getPath("/images/get", &opts)
+		exportpath, err = c.getPath("/images/get", &opts)
 	}
 	if err != nil {
 		return err
 	}
-	return c.streamURL(http.MethodGet, exporturl, streamOptions{
+	return c.stream(http.MethodGet, exportpath, streamOptions{
```
fsouza, line 444: `opts.Context` never reaches the HTTP request: it's copied into the inner options struct (where it has no effect on the query string) but not set as `context:` in these `streamOptions`, so `stream` substitutes `context.Background()` and a cancelled/deadline context can't abort a hung export. Pre-existing omission, but since this function is being rewritten anyway, worth adding `context: opts.Context` here.

client.go
```diff
@@ -866,6 +906,9 @@ func (c *Client) getURL(path string) string {
 	if c.endpointURL.Scheme == unixProtocol || c.endpointURL.Scheme == namedPipeProtocol {
 		urlStr = ""
 	}
+	if expected := c.expectedAPIVersion.Load(); expected != nil {
```
fsouza, line 909: The effective-version precedence rule (`expectedAPIVersion`, else `requestedAPIVersion`) is now hand-copied in `getURL`, `getFakeNativeURL`, and `ExportImages`, while `pathVersionCheck` implements a fourth variant that checks `requestedAPIVersion` first. They agree today only because `setExpectedVersion` mirrors `requestedAPIVersion` verbatim — the moment negotiation gets smarter (e.g. clamping a requested version to the server's max, as moby does), all four sites must change in lockstep or validation and URL-building diverge. A single `c.effectiveAPIVersion()` accessor used everywhere would remove that risk.

client.go
```diff
@@ -441,9 +490,8 @@ func (c *Client) do(method, path string, doOptions doOptions) (*http.Response, e
 		}
 		params = bytes.NewBuffer(buf)
 	}
-	if path != "/version" && !c.SkipServerVersionCheck && c.expectedAPIVersion == nil {
-		err := c.checkAPIVersion()
-		if err != nil {
+	if path != "/version" {
```
fsouza, line 493: This literal `path != "/version"` guard (repeated in `do`, `stream`, and `hijack`) is now the only thing preventing the probe's own `do("/version", ...)` from re-entering `probeServerVersion` while the non-reentrant `versionMu` is held — i.e. it's a deadlock guard, but nothing documents that. Any future change that makes the probe path not byte-equal to `"/version"` (switching to `/_ping`, adding a query param) silently self-deadlocks the first call on a fresh client. An explicit `doOptions` flag (e.g. `skipVersionNegotiation`) set by `getServerAPIVersionString` would encode the invariant where it belongs.

exec.go
```diff
@@ -44,13 +44,12 @@ type CreateExecOptions struct {
 //
 // See https://goo.gl/60TeBP for more details
 func (c *Client) CreateExec(opts CreateExecOptions) (*Exec, error) {
-	if c.serverAPIVersion == nil {
-		c.checkAPIVersion()
-	}
-	if len(opts.Env) > 0 && c.serverAPIVersion.LessThan(apiVersion125) {
+	c.probeServerVersion()
```
fsouza, line 47: Best-effort probing is scattered as bare `c.probeServerVersion()` calls with ignored errors here and in `container_copy.go`, `container_start.go`, and `image.go` (`versionedAuthConfigs`), and each site has a different nil-version policy: exec treats nil as "too old, hard error", container_copy proceeds, container_start picks the new wire format. Also note `probeServerVersion` itself never checks `SkipServerVersionCheck`, so default `NewClient` users now pay a real `/version` round-trip on every one of these calls while the endpoint is failing (nothing is negatively cached), serialized under `versionMu`. A single helper expressing the best-effort contract — including the skip check and ideally failure caching — would fix both.

image.go
```diff
@@ -215,7 +215,11 @@ func (c *Client) InspectImage(name string) (*Image, error) {
 	var image Image
 
 	// if the caller elected to skip checking the server's version, assume it's the latest
-	if c.SkipServerVersionCheck || c.expectedAPIVersion.GreaterThanOrEqualTo(apiVersion112) {
+	if c.SkipServerVersionCheck {
```
fsouza, line 218: The `SkipServerVersionCheck` branch and the `>= 1.12` branch below have identical bodies (decode `resp.Body` into `image`); the original expressed this once behind a combined condition. Something like `v := c.expectedAPIVersion.Load(); if c.SkipServerVersionCheck || (v != nil && v.GreaterThanOrEqualTo(apiVersion112)) { ...decode... } else { ...legacy re-request... }` keeps the semantics without the duplication.

client.go
```diff
@@ -358,20 +362,65 @@ func (c *Client) SetTimeout(t time.Duration) {
...
+func (c *Client) ensureAPIVersion() error {
+	if c.SkipServerVersionCheck || c.expectedAPIVersion.Load() != nil {
+		return nil
+	}
+	return c.probeServerVersion()
+}
+
+// setExpectedVersion records the version used in request URLs. It may differ
+// from the server version when the caller requested an explicit API version.
+func (c *Client) setExpectedVersion(serverVersion APIVersion) {
```
fsouza, line 390: Minor structural point: `setExpectedVersion` re-checks the exact guard `ensureAPIVersion` just evaluated and is called from three separate points inside `probeServerVersion`, so a reader has to trace all three to learn which call actually stores. Simpler shape: `probeServerVersion` only caches `serverAPIVersion`, and `ensureAPIVersion` performs the single expected-version store after a successful probe. (Similarly, both success branches of `pathVersionCheck` end in the identical `Sprintf` — that can collapse to one shared return after the error checks.)

fsouza (commented):
Thanks for working on this — the approach is right and it does fix #1030 (the `ver`-tag minimum versions no longer leak into request URLs, and unversioned clients negotiate the server's version instead). A few issues to sort out before this can merge, though. The first three inline comments are the substantive ones (a behavior regression in `ExportImages`, a silent-failure regression for default `NewClient` users, and a hang risk in the new probe locking); the fourth is an API-compatibility question; the rest are smaller cleanups.

